PR #15, opened 2018-06-01
[CORE-1439] Stream Client code is not transpiled

Aapzu:


test/integration/StreamrClient.spec.js
```diff
@@ -22,6 +24,16 @@ describe('StreamrClient', function () {
         opts = Object.assign({}, defaultOptions, opts)
         return new StreamrClient(opts)
     }
+    
+    before(() => {
+        return fetch('http://localhost:8081')
+            .catch((e) => {
+                if (e.errno === 'ENOTFOUND') {
+                    throw new Error('Integration testing requires the api and http-api ("entire stack") to be run in the background. ' +
```
hpihkala: - Use correct project names: "api and http-api" => "engine-and-editor and data-api"
- Could also check that http://localhost:8890 (data-api) responds

package.json
```diff
@@ -6,13 +6,13 @@
     "type": "git",
     "url": "git://github.com/streamr-dev/streamr-client.git"
   },
-  "main": "src/index.js",
+  "main": "dist/index.js",
```
hpihkala: Filename is probably not `index.js`?

package.json
```diff
@@ -6,13 +6,13 @@
...
   "directories": {
     "test": "test"
   },
   "scripts": {
-    "build": "webpack --env dev && webpack --env prod && npm run test",
-    "dev": "webpack --progress --colors --watch --env dev",
+    "build": "NODE_ENV=production webpack",
```
hpihkala: Previously, `npm build` enforced running tests. I think it's a pretty justifiable idea. For reference, many other build systems, for example `gradle` and `maven` always run tests after building too.

package.json
Aapzu: It's not. Good catch!

package.json
Aapzu: Previously tests were run but not enforced (code was built no matter what the result of the tests were). It was a bug though, and not relevant now.
I personally don't think that failing tests should prevent building, but publishing. I'd rather put the running of tests into eg. pre-commit hook (now when we don't have a CI yet).
If you want to put it back to this command though, we can ofc do it before setting up a proper CI

Aapzu:
@hpihkala pls test again now. Fixed the examples (and tidied them up a bit)

hpihkala:
I noticed that the file size of the `.min.js` had grown from 64kB in `v0.10.1` to 143kB in `v0.10.2`. I managed to track down why. The reason was that the `http` and `https` modules were being webpacked, but they are only used in node to set keepalive for the http agent. I added shims for them, bringing the file size back down to ~68kB minified.

Also did some minor cleanup. All the examples are working now. LGTM. To be published as `v0.11.0`.

Aapzu:
@hpihkala I'll make a pre-release first (`v0.11.0-beta.1`) and start to use that in `marketplace` to be sure that it really works. then later this week we can publish it as a proper version

PR #17, opened 2018-06-02
Update documentation examples to use es6 syntax

Aapzu:
More street cred for using es6😎

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Here are some quick examples. More detailed examples for the browser and node.js
 #### Creating a StreamrClient instance
 
 ```javascript
-var client = new StreamrClient({
+const client = new StreamrClient({
     // See below for more options
     apiKey: 'your-api-key'
 })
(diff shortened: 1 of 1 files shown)


hpihkala:
Hmm - why give an example that has different compatibility level as the distributed code?

I mean, we transpile our es6 code to es5 so that it would work on more browsers, but then we give a usage example in es6? Isn't that a contradiction?

Aapzu:
@hpihkala in a way that's true, of course. But `let`, `const` and arrow functions are such basic features (work in all browsers except IE) that I'd see that not using them is much worse (gives the impression that we are fossil ourselves) than using them even though the code wouldn't work directly in IE anymore

README.md
```diff
@@ -163,14 +163,14 @@ The client and the subscriptions can fire events as detailed below. You can bind
...
 	    console.log('Yeah, we are connected now!')
 	})
 
     // So does the Subscription object
-	var sub = client.subscribe(...)
-	sub.on('subscribed', function() {
-		console.log('Subscribed to '+sub.streamId)
+	const sub = client.subscribe(...)
+	sub.on('subscribed', () => {
+	    console.log('Subscribed to '+sub.streamId)
```
hpihkala: Template literal

hpihkala:
I'm ok with updating examples to something that fail on IE but work in modern browsers. This seems to be the case for const/let, arrow functions and template literals.

README.md
Aapzu: Oops, didn't notice that one

Aapzu:
@hpihkala updated the forgotten template literal

PR #18, opened 2018-06-11
Change from fetch to axios and refactor a bit

Aapzu:
This PR introduces changes, which make adding more domain objects and more API endpoint wrapper easier

- Remove `node-fetch` and introduce `axios`
  - Get rid of content parsing, error management and http/https agent hacks
- Move all API calls from domain object `Stream` to `StreamEndpoints.js`
- Some other, small refactorings

### DIFF
--- a/src/rest/StreamEndpoints.js
+++ b/src/rest/StreamEndpoints.js
@@ -1,5 +1,7 @@
 import qs from 'querystring'
 import debugFactory from 'debug'
+import http from 'http'
+import https from 'https'
 import { authFetch } from './utils'
 
 import Stream from './domain/Stream'
@@ -93,7 +95,7 @@ export function detectStreamFields(streamId, apiKey = this.options.apiKey) {
     return authFetch(`${this.options.restUrl}/streams/${streamId}/detectFields`, apiKey)
 }
 
-export function produceToStream(streamOrId, data, apiKey = this.options.apiKey, requestOptions = {}) {
+export function produceToStream(streamOrId, data, apiKey = this.options.apiKey, requestOptions = {}, keepAlive = true) {
     let streamId
     if (streamOrId instanceof String || typeof streamOrId === 'string') {
         streamId = streamOrId
@@ -106,6 +108,12 @@ export function produceToStream(streamOrId, data, apiKey = this.options.apiKey, 
         `${this.options.restUrl}/streams/${streamId}/data`,
         apiKey,
         {
+            httpAgent: http.Agent({
+                keepAlive,
+            }),
+            httpsAgent: https.Agent({
+                keepAlive,
+            }),
             ...requestOptions,
             method: 'POST',
             data,
--- a/src/rest/StreamEndpoints.js
+++ b/src/rest/StreamEndpoints.js
@@ -1,7 +1,7 @@
 import qs from 'querystring'
-import debugFactory from 'debug'
 import http from 'http'
 import https from 'https'
+import debugFactory from 'debug'
 import { authFetch } from './utils'
 
 import Stream from './domain/Stream'
@@ -23,57 +23,57 @@ export async function listStreams(query = {}, apiKey = this.options.apiKey) {
 }
 
 export async function getStreamByName(name, apiKey = this.options.apiKey) {
-    const json = await this.listStreams({
+    const streamJson = await this.listStreams({
         name,
         public: false,
     }, apiKey)
-    return json[0] ? new Stream(this, json[0]) : undefined
+    return streamJson[0] ? new Stream(this, streamJson[0]) : undefined
 }
 
-export async function createStream(props, apiKey = this.options.apiKey) {
-    if (!props || !props.name) {
+export async function createStream(streamData, apiKey = this.options.apiKey) {
+    if (!streamData || !streamData.name) {
         throw new Error('Stream properties must contain a "name" field!')
     }
 
-    const json = await authFetch(
+    const streamJson = await authFetch(
         `${this.options.restUrl}/streams`,
         apiKey,
         {
             method: 'POST',
-            data: props,
+            data: streamData,
         },
     )
-    return json ? new Stream(this, json) : undefined
+    return streamJson ? new Stream(this, streamJson) : undefined
 }
 
-export async function getOrCreateStream(props, apiKey = this.options.apiKey) {
+export async function getOrCreateStream(streamData, apiKey = this.options.apiKey) {
     let stream
     // Try looking up the stream by id or name, whichever is defined
-    if (props.id) {
-        stream = await this.getStream(props.id, apiKey)
-    } else if (props.name) {
-        stream = await this.getStreamByName(props.name, apiKey)
+    if (streamData.id) {
+        stream = await this.getStream(streamData.id, apiKey)
+    } else if (streamData.name) {
+        stream = await this.getStreamByName(streamData.name, apiKey)
     }
 
     // If not found, try creating the stream
     if (!stream) {
-        stream = await this.createStream(props, apiKey)
-        debug('Created stream: %s (%s)', props.name, stream && stream.id)
+        stream = await this.createStream(streamData, apiKey)
+        debug('Created stream: %s (%s)', streamData.name, stream && stream.id)
     }
 
     return stream
 }
 
-export async function updateStream(streamId, props, apiKey = this.options.apiKey) {
-    const json = await authFetch(
+export async function updateStream(streamId, streamData, apiKey = this.options.apiKey) {
+    const newStreamJson = await authFetch(
         `${this.options.restUrl}/streams/${streamId}`,
         apiKey,
         {
             method: 'PUT',
-            data: props,
+            data: streamData,
         },
     )
-    return json ? new Stream(this, json) : undefined
+    return newStreamJson ? new Stream(this, newStreamJson) : undefined
 }
 
 export function deleteStream(streamId, apiKey = this.options.apiKey) {
--- a/src/rest/StreamEndpoints.js
+++ b/src/rest/StreamEndpoints.js
@@ -1,45 +1,30 @@
 import qs from 'querystring'
-import { Agent as HttpAgent } from 'http'
-import { Agent as HttpsAgent } from 'https'
 import debugFactory from 'debug'
-
-import Stream from './domain/Stream'
 import { authFetch } from './utils'
 
+import Stream from './domain/Stream'
+
 const debug = debugFactory('StreamrClient')
-
-const agentSettings = {
-    keepAlive: true,
-    keepAliveMsecs: 5000,
-}
-
-const agentByProtocol = {
-    http: new HttpAgent(agentSettings),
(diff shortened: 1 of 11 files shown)


src/rest/StreamEndpoints.js
```diff
@@ -1,45 +1,30 @@
...
-    throw new Error(`Unknown protocol in URL: ${url}`)
-}
-
 // These function are mixed in to StreamrClient.prototype.
 // In the below functions, 'this' is intended to be the StreamrClient
 export async function getStream(streamId, apiKey = this.options.apiKey) {
     const url = `${this.options.restUrl}/streams/${streamId}`
-    const json = await authFetch(url, apiKey)
-    return json ? new Stream(this, json) : undefined
+    try {
```
hpihkala: Why this special handling for 404? I think the `authFetch` (old and new version) throws if anything goes wrong (including 404), and that's an ok pattern to keep.

Maybe the answer to "Should an exception be thrown if a resource is not found?" is a matter of opinion, and certainly depends on whether the user of the function expects the resource to be found. For `getStream` he certainly does, as he passes an unguessable id to it => not finding the stream would certainly be an unexpected error situation.

(In the old code, I'm not sure under what conditions `json` could be falsey. Maybe it's a mistake.)

src/rest/StreamEndpoints.js
```diff
@@ -60,52 +45,70 @@ export async function createStream(props, apiKey = this.options.apiKey) {
...
 export async function getOrCreateStream(props, apiKey = this.options.apiKey) {
-    let json
-
-    // Try looking up the stream by id or name, whichever is defined
-    if (props.id) {
-        json = await this.getStream(props.id, apiKey)
-    } else if (props.name) {
-        json = await this.getStreamByName(props.name, apiKey)
-    }
+    let stream = await (props.id ? this.getStream(props.id, apiKey) : this.getStreamByName(props.name, apiKey))
```
hpihkala: Is this code really more readable?

Also it is not equivalent to the old code (new code queries `getStreamByName` even if `props.name` is not defined, which doesn't make sense).

src/rest/StreamEndpoints.js
```diff
@@ -60,52 +45,70 @@ export async function createStream(props, apiKey = this.options.apiKey) {
...
+        stream = await this.createStream(props, apiKey)
     }
 
-    // If still nothing, throw
-    if (!json) {
-        throw new Error(`Unable to find or create stream: ${props.name}`)
-    } else {
-        return new Stream(this, json)
-    }
+    debug('Created stream: %s (%s)', props.name, stream.id)
```
hpihkala: This message should only be shown if a stream was really created (not fetched), just like before.

src/rest/StreamEndpoints.js
```diff
@@ -60,52 +45,70 @@ export async function createStream(props, apiKey = this.options.apiKey) {
...
-    if (!json) {
-        json = await this.createStream(props, apiKey)
-        debug('Created stream: %s (%s)', props.name, json.id)
+    if (!stream) {
+        stream = await this.createStream(props, apiKey)
     }
 
-    // If still nothing, throw
-    if (!json) {
-        throw new Error(`Unable to find or create stream: ${props.name}`)
```
hpihkala: Why not throw on "not found" anymore?

src/rest/utils.js
```diff
@@ -8,24 +8,18 @@ export const authFetch = async (url, apiKey, opts = {}) => {
...
-        }
-    } else if (res.ok) {
-        return {}
-    } else {
-        throw new Error(`Request to ${url} returned with error code ${res.status}: ${text}`)
+    try {
+        const { data } = await axios(url, req)
+        return data
+    } catch (e) {
+        // TODO: what to log here?
```
hpihkala: What about this TODO? I recommend logging (using `debug`) exactly what happened, similar to the old error message of `Request to ${url} returned with error code ${res.status}: ${text}`

src/rest/utils.js
```diff
@@ -8,24 +8,18 @@ export const authFetch = async (url, apiKey, opts = {}) => {
...
+        throw new Error(`${e.message} ${JSON.stringify(e.response.data)}`)
```
hpihkala: Definitely need to include the status code and the failed URL in the error message (like before).

src/rest/StreamEndpoints.js
Aapzu: I guess I got confused because of the checking `return json ? new Stream(this, json) : undefined`

src/rest/StreamEndpoints.js
```diff
@@ -1,45 +1,30 @@
...
-import { Agent as HttpsAgent } from 'https'
 import debugFactory from 'debug'
+import { authFetch } from './utils'
 
 import Stream from './domain/Stream'
-import { authFetch } from './utils'
 
 const debug = debugFactory('StreamrClient')
 
-const agentSettings = {
```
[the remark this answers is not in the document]
Aapzu: Removed them because of a misunderstanding. Will add them back

src/rest/StreamEndpoints.js
Aapzu: `getStream` and `createStream` both throw errors if they do not succeed. If we have got this far we already know we got the response

src/rest/StreamEndpoints.js
Aapzu: Also got confused because inside `getOrCreateStream` the call to `getStream` is not inside `try ... catch` (`client.getOrCreateStream({ id: 'foo', name: 'foo' })` will now throw an error) https://github.com/streamr-dev/streamr-client/pull/18/files#diff-759bd3eca5b1471aff3290ec566e0ed2L74

src/rest/StreamEndpoints.js
```diff
@@ -60,52 +40,82 @@ export async function createStream(props, apiKey = this.options.apiKey) {
...
+
+export function detectStreamFields(streamId, apiKey = this.options.apiKey) {
+    return authFetch(`${this.options.restUrl}/streams/${streamId}/detectFields`, apiKey)
+}
+
+export function produceToStream(streamOrId, data, apiKey = this.options.apiKey, requestOptions = {}, keepAlive = true) {
     let streamId
-    if (streamObjectOrId instanceof Stream) {
-        streamId = streamObjectOrId.id
+    if (streamOrId instanceof String || typeof streamOrId === 'string') {
```
timoxley: Should this be `streamOrId instanceof Stream` rather than `String`?

timoxley: Oh I see, yeah IMO I wouldn't even try to handle the `new String()` scenario, those types of values are going to be bad news for almost everything doing any kind of argument checks. `typeof` only IMO.

src/rest/domain/GenericDomainObject.js
```diff
@@ -0,0 +1,13 @@
+export default class GenericDomainObject {
+    constructor(client, props = {}) {
+        if (!client) {
+            throw new Error('A client must be given')
+        }
+        this._client = client
+        Object.assign(this, props)
+    }
+
+    setProps(props = {}) {
```
timoxley: Maybe `setState` for parity with React's Component API?

src/rest/StreamEndpoints.js
Aapzu: Good to know, I just have a habit of checking a string that way (from some of the earliest "how to check if a value is a string in javascript" google searches)

src/rest/utils.js
```diff
@@ -8,24 +8,17 @@ export const authFetch = async (url, apiKey, opts = {}) => {
...
-        }
-    } else if (res.ok) {
-        return {}
-    } else {
-        throw new Error(`Request to ${url} returned with error code ${res.status}: ${text}`)
+    try {
+        const { data } = await axios(url, req)
+        return data
+    } catch (e) {
+        throw new Error(`Request to ${url} failed with http status code ${e.response.status}: ${JSON.stringify(e.response.data)}`)
```
timoxley: This crash handler is going to crash if `e.response` or `e.response.data` don't exist for whatever reason.

test/integration/StreamrClient.spec.js
```diff
@@ -22,16 +22,14 @@ describe('StreamrClient', function () {
...
 
     before(() => Promise.all([ // TODO: Figure out how to handle this when setting up a CI
-        fetch(`http://${engineAndEditor}`),
-        fetch(`http://${dataApi}`),
+        axios.get(`http://${engineAndEditor}`),
+        axios.get(`http://${dataApi}`),
     ])
         .catch((e) => {
-            if (e.errno === 'ENOTFOUND' || e.errno === 'ECONNREFUSED') {
+            if (e.code === 'ENOTFOUND' || e.code === 'ECONNREFUSED') {
```
timoxley: are you sure these codes are right? when I run:

```
axios.get('http://httpstat.us/404').then(() => console.log('ok'), (e) => console.error(e.code)
```

the code is `undefined`.

src/rest/domain/GenericDomainObject.js
Aapzu: That's a good point. In a way I'd like to use some completely different word than `props` or `state`, but definitely from those two `state` is better

src/rest/utils.js
Aapzu: @hpihkala wanted the exception to show the url and the original exception doesn't show that

test/integration/StreamrClient.spec.js
Aapzu: At least when running the code in Node without `data-api` running in the background there is. But you're correct that when running in browser it's hard to get it to show the field (=didn't succeed)

Aapzu:
@hpihkala, added `httpAgent` and `httpsAgent` options back. Changed the error message of `authFetch` to include the url. Removed the special handling of 404 in `getStream`.
One more question: what do we want 
```
client.getOrCreateStream({
  id: 'foobar',
  name: 'name',
})
```
to do in a case where a Stream with id `foobar` does not exist?

hpihkala:
@Aapzu giving an `id` to `getOrCreateStream` doesn't make sense. This could be blocked already in the client by checking if the given object has an `id` field, and if it does, throwing an error. (The error could instruct the user to use `getStream` for getting streams by id).

src/rest/StreamEndpoints.js
```diff
@@ -106,6 +108,12 @@ export function produceToStream(streamOrId, data, apiKey = this.options.apiKey,
         `${this.options.restUrl}/streams/${streamId}/data`,
         apiKey,
         {
+            httpAgent: http.Agent({
```
hpihkala: A few questions regarding this fix:

- Is this creating a new `Agent` for every request? (Here, actually **two** agents for every request). I would assume that the same `Agent` is meant to be shared across many requests.

- I'm not sure what exactly this line does. [The docs](https://nodejs.org/api/http.html#http_class_http_agent) show agent creation which includes the `new` keyword.

hpihkala:
Closing old unfinished PR

PR #21, opened 2018-09-19
Recover after messages that fail to parse

hpihkala:
The `lastReceivedOffset` variable in `Subscription` is now updated after an `InvalidJsonError` is encountered to avoid doomed-to-fail resends triggered by the gapfill protocol. This allows the flow of messages to continue normally after the message that failed to parse.

### DIFF
--- a/src/Protocol.js
+++ b/src/Protocol.js
@@ -13,6 +13,11 @@ messageTypesByCode.forEach((type, idx) => {
 
 const BYE_KEY = '_bye'
 
+// Slow, use only in exceptional situations
+function getField(fieldName, msg) {
+    return msg[fieldsByProtocolVersion[msg[0]].indexOf(fieldName)]
+}
+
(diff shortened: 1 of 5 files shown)


hpihkala:
@Aapzu do you have time to review this? This is the thing that you started, we had a chat about it, then I continued it.

src/Protocol.js
```diff
@@ -13,6 +13,11 @@ messageTypesByCode.forEach((type, idx) => {
 
 const BYE_KEY = '_bye'
 
+// Slow, use only in exceptional situations
+function getField(fieldName, msg) {
```
Aapzu: More es6 way would be 
```
const getField(fieldName, msg) => msg[fieldsByProtocolVersion[msg[0]].indexOf(fieldName)]
```
but it really doesn't matter:)

PR #24, opened 2018-09-27
Build in npm prepare step

hpihkala:
@Aapzu Cool! Like this?


mondoreale:
Missed the message, sorry. Yep, just like in `streamr-layout`. 👍

PR #26, opened 2018-10-31
Core 1016 authentication project

mthambipillai:


src/rest/utils.js
```diff
@@ -3,16 +3,18 @@ import debugFactory from 'debug'
 
 const debug = debugFactory('StreamrClient:utils')
 
-export const authFetch = async (url, apiKey, opts = {}) => {
+export const authFetch = async (url, session, opts = {}, requireNewToken = false) => {
     debug('authFetch: ', url, opts)
+    const newHeaders = opts.headers ? opts.headers : {}
+    if (session) {
+        const token = await session.getSessionToken(requireNewToken)
+        newHeaders.Authorization = `Bearer ${token}`
```
hpihkala: Here you're modifying the `opts.headers` object which is potentially passed from outside this function, which is bad practice as the function effects leak outside the function scope. One idea would be to do something like (untested code):

```
const authHeader = {}
if (session) {
    const token = await session.getSessionToken(requireNewToken)
    authHeader.Authorization = `Bearer ${token}`
}

const req = {
    ...opts,
    headers: {
        ...authHeader,
        ...(opts.headers || {})
    }
}
```

test/integration/LoginEndpoints.test.js
```diff
@@ -0,0 +1,127 @@
...
+        })
+            .then((challenge) => {
+                assert(challenge)
+                assert(challenge.id)
+                assert(challenge.challenge)
+                assert(challenge.expires)
+            }))
+    })
+
+    async function assertThrowsAsync(fn, regExp) {
```
hpihkala: Hmm. Surely there's a built-in and elegant way to test that a promise is rejected? For such a common case you shouldn't need to roll your own?

test/integration/LoginEndpoints.test.js
```diff
@@ -0,0 +1,127 @@
...
+                    id: 'some-id',
+                    challenge: 'some-challenge',
+                },
+                signature: 'some-sig',
+                address: 'some-address',
+            }), /Error/)
+        })
+        it('login should pass and should receive a session token', () => {
+            const account = web3.eth.accounts.create()
+            client.getChallenge({
```
hpihkala: Hmm you probably want to `return` this whole thing (the promise), otherwise the test will always succeed and move on to the next one before finishing!

test/integration/LoginEndpoints.test.js
```diff
@@ -0,0 +1,127 @@
...
+                        .then((sessionToken) => {
+                            assert(sessionToken)
+                            assert(sessionToken.token)
+                            assert(sessionToken.expires)
+                        })
+                })
+        })
+        it('login should pass using combined function', () => {
+            const account = web3.eth.accounts.create()
+            client.loginWithChallengeResponse((d) => account.sign(d).signature, account.address)
```
hpihkala: I think `return` here too

test/integration/LoginEndpoints.test.js
```diff
@@ -0,0 +1,127 @@
+import 'babel-polyfill' // Needed because of mocha
```
hpihkala: Actually we might not need this import anymore (we're not even using `mocha` anymore). Could you check if everything works without it? If yes, please remove it from other integration test files too! 👍

test/integration/LoginEndpoints.test.js
```diff
@@ -0,0 +1,127 @@
...
+            f = () => {
+                throw e
+            }
+        } finally {
+            assert.throws(f, regExp)
+        }
+    }
+
+    describe('Challenge response', () => {
+        it('login should fail', async () => {
```
hpihkala: Minor thing, but the point is to create sentences beginning with "it" ("it" refers to what is being `describe`d):

_it should fail_ (in other words, _Challenge response should fail_)

whereas the below sentence doesn't work: 

_it login should fail_

test/integration/StreamEndpoints.test.js
```diff
@@ -36,6 +52,35 @@ describe('StreamEndpoints', () => {
                 assert.equal(stream.name, name)
             }))
 
+        it('createStream with private key', () => clientPrivateKey.createStream({
```
hpihkala: We don't want to repeat every endpoint test with every authentication method, but then again we want to test the automatic login. How to accomplish that and where should these tests go?

Let's start from figuring out what class contains the code we're testing. In this case, I guess it's actually in `utils.js` (in `authFetch`, right? Could we write a test for that? Maybe it's a unit test even, as we can pass a mock `session` to it?

test/integration/StreamrClient.test.js
```diff
@@ -16,7 +16,9 @@ describe('StreamrClient', () => {
     const createClient = (opts = {}) => new StreamrClient({
         url: config.websocketUrl,
         restUrl: config.restUrl,
-        apiKey: 'tester1-api-key',
+        auth: {
+            privateKey: '0x12345564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709',
```
hpihkala: Side note: unlike addresses, the private key is often expressed without `0x` prefix. I wonder if that works too?

src/StreamrClient.js
```diff
@@ -245,7 +259,7 @@ export default class StreamrClient extends EventEmitter {
         }
     }
 
-    subscribe(optionsOrStreamId, callback, legacyOptions) {
+    async subscribe(optionsOrStreamId, callback, legacyOptions) {
```
hpihkala: This completely breaks existing clients! :( Ideally `subscribe` would keep returning the `Subscription` instead of a `Promise` (which `async` will do). It should be possible if we don't needlessly call `getSessionToken()` here and pass it to the `Subscription` (see my comments in `Subscription`)

test/unit/StreamrClient.test.js
```diff
@@ -50,10 +50,10 @@ describe('StreamrClient', () => {
         asyncs = []
     }
 
-    function setupSubscription(streamId, emitSubscribed = true, subscribeOptions = {}, handler = sinon.stub()) {
+    async function setupSubscription(streamId, emitSubscribed = true, subscribeOptions = {}, handler = sinon.stub()) {
         assert(client.isConnected(), 'setupSubscription: Client is not connected!')
         connection.expect(new SubscribeRequest(streamId))
-        const sub = client.subscribe({
+        const sub = await client.subscribe({
```
hpihkala: See comment on `subscribe` - we should keep returning the `Subscription` to avoid breaking things. Just commenting here to detect the change

test/integration/LoginEndpoints.test.js
mthambipillai: Yes there is `assert.rejects()` in Node 10, but we use Node 8... I didn't really create this function, I found it online as a workaround for versions prior to Node 10. I don't know if there exists a more elegant way to do it.

test/integration/StreamEndpoints.test.js
mthambipillai: Could we say that all integration tests which are related to authentication should be in `Session.test.js`? This way I could remove the additional clients created in this file and test some of these endpoints in `Session.test.js` with the different existing clients in that file.

src/Session.js
```diff
@@ -48,8 +34,20 @@ export default class Session {
         if (this.options.sessionToken && !requireNewToken) {
             return this.options.sessionToken
         }
-        const tokenObj = await this.loginFunction()
-        this.options.sessionToken = tokenObj.token
-        return tokenObj.token
+        if (this.state === Session.State.LOGGING_IN) {
+            return Promise.reject(new Error('Already logging in!'))
```
hpihkala: Multiple `getSessionToken` calls would now automatically fail except the first one. This is a common situation when for example data produce requests occur in quick succession before the `loginFunction` completes. I think the functionality we want is this (this test case should be added):

- call `getSessionToken()` (returns `promise1`)
- assert `loginFunction` called once
- call `getSessionToken()` again (returns `promise2`)
- assert `loginFunction` still called once
- `loginFunction` resolves
- use `Promise.all` to combine `promise1` and `promise2`, then in `then` block check that both promises resolved with the **same** session token

Another test case (login failure):

- call `getSessionToken()` (returns `promise1`)
- assert `loginFunction` called once
- call `getSessionToken()` again (returns `promise2`)
- assert `loginFunction` still called once
- `loginFunction` **rejects**
- both promises should reject

Ok, so how to implement this behavior? How to make the second (and third, etc.) promise to resolve in a delayed fashion?

Two approaches come to mind:

**1. Return the original promise**

Save the Promise returned by `loginFunction` into some instance field and return that same Promise for every subsequent call if the state is `LOGGING_IN`. This would be an elegant solution but I'm not 100% sure if it works in practice.

**2. Wait for state change**

If we are `LOGGING_IN`, instead of the current auto-rejection, return a Promise that e.g. periodically checks if the state has changed to `LOGGED_IN`, and if it has, resolve the promise with the `sessionToken`. If the state has changed to `LOGGED_OUT` (you should add a `catch` block to the `loginFunction` call to set this state), reject the promise. Something like (completely untested):

```
if (this.state === Session.State.LOGGING_IN) {
    return new Promise((resolve, reject) => {
        let attempt = 0
        const check = () => {
            if (this.state === Session.State.LOGGED_IN) {
                resolve(this.options.sessionToken)
            } else if (this.state === Session.State.LOGGED_OUT || attempt > MAX) {
                reject(new Error('Was waiting for login, but it failed or timed out!'))
            } else {
                attempt += 1
                setTimeout(check, 100)
            }
        }
        setTimeout(check, 100)
    }
}
```

src/StreamrClient.js
```diff
@@ -29,33 +29,18 @@ export default class StreamrClient extends EventEmitter {
...
-        this.session = new Session(this)
-
-        // Backwards compatibility for the use of apiKey in code other than strict authentication by the Session
-        this.options.apiKey = this.options.auth.apiKey
-
-        // Backwards compatibility for option 'authKey' => 'apiKey'
-        if (this.options.authKey && !this.options.apiKey) {
-            this.options.apiKey = this.options.authKey
-        }
+        this.session = new Session(this, options)
```
hpihkala: Why not pass `options.auth` to `Session`? It only reads that part and discards the rest, right?

test/integration/Session.test.js
```diff
@@ -37,17 +49,27 @@ describe('Session', () => {
...
                 assert(sessionToken)
             }))
     })
+
+    describe('Internal state', () => {
+        it('should throw when calling getSessionToken() while logging in', async () => {
+            clientApiKey.session.options.sessionToken = undefined
+            await assertThrowsAsync(async () => Promise.all([
+                clientApiKey.session.getSessionToken(),
+                clientApiKey.session.getSessionToken(),
```
hpihkala: See discussion related to this in `Session`

test/integration/StreamEndpoints.test.js
hpihkala: Well basically there's 2 things happening that should get tested:

1. `Session` can accept different kinds of authentication (username/password, apiKey, sessionToken). That functionality (and hence those tests cases) should be in `Session.test.js` (and IIRC they are already!).

2. `authFetch` asks for the token from `session` and uses it in the request. If the request returns 400 or 401, it forces a new token and retries. This functionality is in `utils.js` so we should have a `utils.test.js` that tests it.

If you look at the code in `StreamEndpoints`, it mainly just passes the `session` object around, so that code doesn't care about what type of credentials are contained within the session object. So I think we can ignore the different credential types in the endpoint tests (just pick one), and instead make sure functionalities 1. and 2. are covered in respective test classes.

src/StreamrClient.js
```diff
@@ -28,18 +29,15 @@ export default class StreamrClient extends EventEmitter {
...
         }
         this.subsByStream = {}
         this.subById = {}
         this.publishQueue = []
 
         Object.assign(this.options, options || {})
 
-        // Backwards compatibility for option 'authKey' => 'apiKey'
-        if (this.options.authKey && !this.options.apiKey) {
-            this.options.apiKey = this.options.authKey
```
hpihkala: Whoops, was the backwards compatibility now accidentally dropped? Please add two test cases to ensure it works:

1. test `new StreamrClient({ authKey: 'foo' })` results in `this.options.auth.apiKey === 'foo'`
2. test `new StreamrClient({ apiKey: 'foo' })` results in `this.options.auth.apiKey === 'foo'`

package.json
```diff
@@ -64,6 +64,7 @@
     "node-fetch": "^2.1.2",
     "querystring": "^0.2.0",
     "streamr-client-protocol": "^1.0.0",
+    "web3": "^1.0.0-beta.36",
```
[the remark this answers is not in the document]
hpihkala: Oh crap, a 10-fold increase in file size is pretty devastating :( 

Let's first prep this branch to a ready state, and maybe you can quickly try out `ethjs` in another branch to see if it works/helps. Btw what do we need `ethjs-query` for?

The [ethjs readme](https://github.com/ethjs/ethjs) brags about being just 100 KB minified, and it must include some combination of functionality from its individual packages. I'm not sure what's going on there.

package.json
```diff
@@ -61,6 +61,7 @@
     "babel-runtime": "^6.26.0",
     "debug": "^3.1.0",
     "eventemitter3": "^3.0.1",
+    "express": "^4.16.4",
```
hpihkala: Should be in `devDependencies`? Only needed in tests, right?

src/Session.js
```diff
@@ -34,11 +34,11 @@ export default class Session {
...
         }
-        if (this.state === Session.State.LOGGING_IN) {
-            return Promise.reject(new Error('Already logging in!'))
+        if (this.state !== Session.State.LOGGING_IN) {
+            this.state = Session.State.LOGGING_IN
+            this.sessionTokenPromise = this.loginFunction()
         }
-        this.state = Session.State.LOGGING_IN
-        return this.loginFunction().then((tokenObj) => {
+        return this.sessionTokenPromise.then((tokenObj) => {
```
hpihkala: I think there's a bug here - now a new `then` handler is added to this single promise for every call to `getSessionToken`! I think you should attach the `then` handler on line 39 where you create the Promise, and on this line just `return this.sessionTokenPromise`.

Please also add a `catch` block that sets `this.state` to `Session.State.LOGGED_OUT` and throws the error onwards. Otherwise if the login attempt throws, the `Session` will be forever stuck to `LOGGING_IN` and subsequent attempts cannot succeed.

src/Session.js
```diff
@@ -36,13 +36,18 @@ export default class Session {
...
         if (this.state !== Session.State.LOGGING_IN) {
             this.state = Session.State.LOGGING_IN
-            this.sessionTokenPromise = this.loginFunction()
+            try {
+                this.sessionTokenPromise = this.loginFunction().then((tokenObj) => {
+                    this.options.sessionToken = tokenObj.token
+                    this.state = Session.State.LOGGED_IN
+                    return tokenObj.token
+                })
+            } catch (err) {
```
hpihkala: This should probably be a `.catch(...)` for the promise, not an actual `try-catch` block, since any errors from `loginFunction` will appear as rejection of the promise, right? 

Before changing this part, please create a test (which fails) first, since I think we're missing coverage on that part. It's a good pattern worth learning anyway: notice bug, create test (should fail!), fix bug, run test (should pass!).

Since there's some non-trivial functionality in `Session.js`, I think it deserves a set of unit tests (in addition to the integration test). There we can easily mock the `_client.loginWith*` functionalities for detailed testing. Plus in the current integration test, the stuff in `describe('Internal state', ...)` would be better off moved to a unit test, allowing you to check that the `loginFunction` is called an expected number of times, as discussed in one of the previous comments.

test/integration/Session.test.js
```diff
@@ -0,0 +1,89 @@
...
+    const createClient = (opts = {}) => new StreamrClient({
+        url: config.websocketUrl,
+        restUrl: config.restUrl,
+        autoConnect: false,
+        autoDisconnect: false,
+        ...opts,
+    })
+
+    beforeAll(() => {
+        clientApiKey = createClient({
```
hpihkala: Missing a test case for incorrect credentials (should lead to `loginFunction` promise rejected)

test/integration/Session.test.js
```diff
@@ -0,0 +1,89 @@
...
+                assert(sessionToken)
+            }))
+        it('should fail to get token with no authentication', (done) => clientNone.session.getSessionToken()
+            .catch((err) => {
+                assert.equal(err.toString(), 'Error: Need either "privateKey", "apiKey" or "username"+"password" to login.')
+                done()
+            }))
+    })
+
+    describe('Internal state', () => {
```
hpihkala: This stuff would be better suited for a unit test (see comment in `Session.js`, and earlier comment about testing the `loginFunction` calls). Then you can mock the collaborator calls (under `_client`) with `sinon.mock()`s, making it very easy to test that `Session` reacts appropriately.

src/Session.js
hpihkala: My latest comments relate to this, so I'll comment here too: I'm not sure what you mean by "private function", as there are no actual private methods in JS? Anyway, I would propose a unit test where calls to outside of `Session` can easily be mocked, allowing you to implement the above outlined test cases.

README.md
```diff
@@ -79,10 +81,21 @@ Option | Default value | Description
 ------ | ------------- | -----------
 url | wss://www.streamr.com/api/v1/ws | Address of the Streamr websocket endpoint to connect to.
 restUrl | https://www.streamr.com/api/v1/ws | Base URL of the Streamr REST API.
-apiKey | null | Define a default apiKey to use when none is specified in the individual function calls.
+apiKey | null | Defined for backwards compatibility. Should use `auth.apiKey` instead.
```
hpihkala: I think we can leave this line out of the readme. Old functionality will still work, but let's not point people to it anymore.

src/Session.js
```diff
@@ -13,12 +13,15 @@ export default class Session {
             this.loginFunction = async () => this._client.loginWithChallengeResponse((d) => account.sign(d).signature, account.address)
         } else if (this.options.provider) {
             const w3 = new Web3(this.options.provider)
-            const accounts = w3.eth.getAccounts()
-            const address = accounts[0]
-            if (!address) {
-                throw new Error('Cannot access account from provider')
+            this.loginFunction = async () => {
+                const address = await w3.eth.getAccounts().then((accounts) => {
```
hpihkala: Unusual mix of `await` and `.then()` here! Pick one and stick with it? With `await` it would simplify to something like this (untested):
```
const accounts = await w3.eth.getAccounts()
if (!accounts[0]) {
    throw new Error('Cannot access account from provider')
}
return this._client.loginWithChallengeResponse((d) => w3.eth.personal.sign(d, accounts[0]), accounts[0])
```

hpihkala:
@mthambipillai `refactor-oom-protocol` branch is now merged to `master` - you can change the base branch in this PR to `master`.

PR #27, opened 2018-11-14
Data integrity project

mthambipillai:


hpihkala:
@mthambipillai You started this branch from your authentication branch (which was absolutely the right thing to do!), just one minor thing: you need to set that as base in this PR! Now the diff is showing me all the authentication changes too, because `refactor-oom-protocol` obviously doesn't have them! That mixup makes this PR impossible to review atm.

As the authentication stuff is already merged to `master` now, please base this PR on `master` and everything should be fine!

src/StreamrClient.js
```diff
@@ -435,10 +437,15 @@ export default class StreamrClient extends EventEmitter {
...
     }
 
-    _requestPublish(streamId, data, apiKey, sessionToken) {
-        const request = new PublishRequest(streamId, apiKey, sessionToken, data)
+    _requestPublish(streamId, data, timestamp, apiKey, sessionToken) {
+        const request = new PublishRequest(streamId, apiKey, sessionToken, data, timestamp)
         debug('_requestResend: %o', request)
-        this.connection.send(request)
+        if (this.signer) {
+            return this.signer.getSignedPublishRequest(request).then((signedRequest) => {
```
hpihkala: To avoid repeating `this.connection.send` (and fix the `debug`) you could:

```
let request = new PublishRequest(streamId, apiKey, sessionToken, data, timestamp)
if (this.signer) {
    request = await this.signer.getSignedPublishRequest(request)
}
debug('_requestPublish: %o', request)
return this.connection.send(request)

File: src/rest/domain/Stream.js [hunk 2]
```diff
@@ -52,7 +52,7 @@ export default class Stream {
         )
     }
 
-    produce(data) {
-        return this._client.produceToStream(this.id, data)
+    produce(data, timestamp = undefined) {
```
hpihkala:
```suggestion
    produce(data, timestamp = Date.now()) {
```

test/integration/Signer.test.js
```diff
@@ -0,0 +1,124 @@
+import assert from 'assert'
```
hpihkala: This test duplicates a lot of code from `StreamrClient.test.js`, especially the setup and teardown stuff. I was just wondering if maybe it would be better after all to add these test cases to `StreamrClient.test.js`?

test/integration/StreamrClient.test.js
```diff
@@ -16,7 +16,7 @@ describe('StreamrClient', () => {
         url: config.websocketUrl,
         restUrl: config.restUrl,
         auth: {
-            privateKey: '12345564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709',
+            privateKey: '0x12345564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709',
```
hpihkala: The private key is often presented also without the `0x` in the beginning. Can we make it work both ways?

src/StreamrClient.js
```diff
@@ -46,6 +47,7 @@ export default class StreamrClient extends EventEmitter {
         }
 
         this.session = new Session(this, options.auth)
+        this.signer = (options.auth && options.auth.publishWithSignature) ? new Signer(options.auth) : undefined
```
hpihkala: The option feels like it's in the wrong place under `auth`. Would be cool to also have care-free default functionality. How about something like this:

```
new StreamrClient({
    auth: {
        // as before
    },
    publishWithSignature: 'auto', // valid values: 'auto' (default), 'always', 'never'
})
```

- 'auto': sign messages if possible (key or provider given)
- 'always': always sign messages (throw if not possible)
- 'never': never sign messages (even if possible)

hpihkala:
Also, please remember to update `README`

src/Signer.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    static verifySignature(data, signature, address, signatureType = SIGNATURE_TYPE_ETH) {
+        if (signatureType === SIGNATURE_TYPE_ETH) {
+            return web3.eth.accounts.recover(data, signature).toLowerCase() === address.toLowerCase()
+        }
+        throw new Error(`Unrecognized signature type: ${signatureType}`)
+    }
+
+    // TODO: should be used by the StreamrClient before calling Subscription.handleMessage but only if client required signature verification
```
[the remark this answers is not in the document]
mthambipillai: I left it as a TODO because I remember that we wanted to make the signature computation work first and we didn't discuss how exactly signature verification should be done.
Right now this method is not used except by tests. The `StreamrClient` would first need to know the set of publishers for every stream.

src/Signer.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+        throw new Error(`Unrecognized signature type: ${signatureType}`)
+    }
+
+    async getSignedPublishRequest(publishRequest, signatureType = SIGNATURE_TYPE_ETH) {
+        const ts = publishRequest.getTimestampAsNumber()
+        if (!ts) {
+            throw new Error('Timestamp is required as part of the data to sign.')
+        }
+        const payload = this.address.toLowerCase() + publishRequest.streamId + ts + publishRequest.getSerializedContent()
```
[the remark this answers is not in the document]
mthambipillai: Yes I think a new `signatureType` in case of a new payload to sign makes sense. I would maintain the `signData` because the signing algorithm itself might change and not the payload.

src/StreamrClient.js
```diff
@@ -46,8 +47,12 @@ export default class StreamrClient extends EventEmitter {
             this.options.auth.apiKey = this.options.apiKey
         }
 
-        this.session = new Session(this, options.auth)
-        this.signer = (options.auth && options.auth.publishWithSignature) ? new Signer(options.auth) : undefined
+        if (this.options.privateKey && !this.options.privateKey.startsWith('0x')) {
```
hpihkala: Bug? Should be `this.options.auth.privateKey`, right? If this is uncaught by tests, then please add a unit test that gives a `privateKey` without `0x` and asserts that it gets added.

test/unit/Signer.test.js
```diff
@@ -66,10 +94,11 @@ describe('Signer', () => {
                 signedRequest.publisherAddress,
                 signedRequest.signature,
             )
-            Signer.verifyStreamMessage(streamMessage)
+            Signer.verifyStreamMessage(streamMessage, new Set([signedRequest.publisherAddress]))
         })
 
         it('Should throw if incorrect signature', async () => {
```
hpihkala: Doesn't need to be `async`?

test/unit/Signer.test.js
```diff
@@ -82,10 +111,17 @@ describe('Signer', () => {
...
                 1,
-                '0xF915eD664e43C50eB7b9Ca7CfEB992703eDe55c4',
+                address,
                 '0x3d5c221ebed6bf75ecd0ca8751aa18401ac60561034e3b2889dfd7bbc0a2ff3c5f1c5239113f3fac5b648ab665d152ecece1daaafdd3d94309c2b822ec28369e1c',
             )
-            assert.throws(() => Signer.verifyStreamMessage(streamMessage), Error)
+            assert.throws(() => Signer.verifyStreamMessage(streamMessage, new Set([address])), Error)
+        })
+
+        it('Should throw if correct signature but not from a trusted publisher', async () => {
```
hpihkala: Doesn't need to be `async`

test/unit/Signer.test.js
```diff
@@ -82,10 +111,17 @@ describe('Signer', () => {
...
+            const streamMessage = new StreamMessage()
+            Signer.getPayloadToSign = sinon.stub().returns('')
```
hpihkala: Whoopsie! You're overwriting the static method on the class. The class will remain crippled and any test case that comes after this one will fail spectacularly.

It might not be a great idea anyway to stub internals of the class under test: better would be the usual pattern of providing valid input to the public method that you're testing (like you're doing in the other tests).

To avoid repeating the code to create a signed message to test, just create one in the `beforeEach` block and use that in the test cases, tweaking it to suit the particular test case in question (mess up the signature etc.).

test/unit/Signer.test.js
mthambipillai: I dont think so because I don't see any `async` function inside `verifyStreamMessage`. `web3.eth.accounts.recover` is not `async`.

test/unit/StreamrClient.test.js
```diff
@@ -321,22 +323,36 @@ describe('StreamrClient', () => {
...
                 connection.expect(new SubscribeRequest('stream1'))
 
                 const counter = sinon.stub()
+                counter.onFirstCall().returns(1)
+                counter.onSecondCall().returns(2)
 
                 client.subscribe({
                     stream: 'stream1',
-                }, counter)
+                }, () => {
```
hpihkala: There must be a simpler way! I can't write it off the top of my head, but I was thinking something like returning a promise that resolves on the second call. Thought about how to do it with sinon stubs easily, but I didn't figure it out immediately. Moving on.

test/unit/StreamrClient.test.js
```diff
@@ -587,6 +603,14 @@ describe('StreamrClient', () => {
                 })
             })
 
+            it('should add a new SubscribedStream', () => {
+                const streamId = 'streamId'
+                connection.expect(new SubscribeRequest(streamId))
+                client.subscribe(streamId, () => {})
+                const stream = client.subscribedStreams.streamId
+                assert(stream instanceof SubscribedStream)
```
hpihkala: Hmm, isn't `stream` a string (`streamId`)? (see previous line)

test/unit/StreamrClient.test.js
```diff
@@ -587,6 +603,14 @@ describe('StreamrClient', () => {
                 })
             })
 
+            it('should add a new SubscribedStream', () => {
+                const streamId = 'streamId'
+                connection.expect(new SubscribeRequest(streamId))
+                client.subscribe(streamId, () => {})
+                const stream = client.subscribedStreams.streamId
```
hpihkala: Digging into module internals isn't a great pattern. Is there publicly exposed functionality we can test instead? Is this already covered by existing tests?

test/unit/SubscribedStream.test.js
```diff
@@ -0,0 +1,144 @@
+import assert from 'assert'
+import sinon from 'sinon'
+import { PublishRequest, StreamMessage } from 'streamr-client-protocol'
+import StreamrClient from '../../src/StreamrClient'
+import SubscribedStream from '../../src/SubscribedStream'
+import Signer from '../../src/Signer'
+
+describe('SubscribedStream', () => {
```
hpihkala: Comment applies to several places in this file: you're testing (and manipulating) implementation _internals_. Is there a way to test _behavior_ instead (using the _public API_ of the class)?

Example of antipattern:
- When `StreamrClient` has option `verifySignatures: 'auto'`, internal flag `verifySignatures` has value `undefined`

Example of behavioral pattern:
- When `StreamrClient` has option `verifySignatures: 'auto'`, message signature is verified if present

How to observe behavior, then?
- Public API
- Collaborator calls (stubbed)

src/SubscribedStream.js
```diff
@@ -17,10 +17,10 @@ export default class SubscribedStream {
...
-        if (!this.producers) {
-            this.producers = await this._client.getStreamProducers(this.streamId)
+        if (!this.producersPromise) {
+            this.producersPromise = this._client.getStreamProducers(this.streamId)
         }
-        return this.producers
+        return this.producersPromise
     }
 
     async verifyStreamMessage(msg) {
```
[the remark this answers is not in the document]
mthambipillai: So I believe the logic you want is the following (in case `client.options.verifySignatures == 'auto'`)?

- message signed AND stream requires signed data --> verify
- message signed AND stream doesn't require signed data --> verify
- message not signed AND stream requires signed data --> false
- message not signed AND stream doesn't require signed data --> true

I'm not sure about the second line, I am wondering if instead it should return 'true' and not verify the signature. Isn't it possible for a malicious node to add a random signature to received messages and forward it to other nodes who will reject it (kind of DoS attack)?
I am also wondering what would be a legitimate case where a signed message is produced on a stream that normally doesn't require signed data.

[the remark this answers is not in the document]
hpihkala: Yes the table above is exactly what I meant.

> Isn't it possible for a malicious node to add a random signature to received messages and forward it to other nodes who will reject it (kind of DoS attack)?

Any malicious node could achieve the same effect by simply dropping the messages instead of going through the trouble of even spoofing a signature. In this case other (honest) nodes would still deliver the valid data. If the malicious node is the entry node, the client can notice that their messages are being censored and switch to another node, or produce to multiple nodes.

> I'm not sure about the second line, I am wondering if instead it should return 'true' and not verify the signature.

Hmm.. What reason would the client have *not* to verify a provided signature? It's extra information which can only benefit the client, right? A malicious node could remove the signature and tamper the data, but client's shouldn't trust unsigned data anyway, and this outcome is not worse than ignoring the signature.

PR #30, opened 2018-11-21
fix: CORE-1568 - Integration tests failing:

mikhaelsantos:
* Add bind ip
* Increment max user watches
* Add code to wait for EE
* Deploy services by parts
* Add jest option 'max-workers'

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+dist: trusty
 node_js:
 - '8'
 sudo: true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,7 +17,15 @@ jobs:
   - stage: Integration tests
     script:
     - sudo /etc/init.d/mysql stop
+    - sudo sysctl fs.inotify.max_user_watches=524288; sudo sysctl -p
     - git clone https://github.com/streamr-dev/streamr-docker-dev.git
+    - sudo ifconfig docker0 10.200.10.1/24
```
hpihkala: Didn't `streamr-docker-dev bind-ip` work? Why duplicate the functionality?

.travis.yml
```diff
@@ -16,7 +17,15 @@ jobs:
...
+    - "$TRAVIS_BUILD_DIR/streamr-docker-dev/streamr-docker-dev/bin.sh start 4"
```
hpihkala: I think there's redundancy here: `streamr-docker-dev start engine-and-editor` will actually start `engine-and-editor` and all its dependencies, including all the services in "4", so I guess this line could just be removed?

.travis.yml
```diff
@@ -16,7 +17,15 @@ jobs:
...
+    - "$TRAVIS_BUILD_DIR/streamr-docker-dev/streamr-docker-dev/bin.sh start engine-and-editor"
+    - while true; do http_code=$(curl -s -o /dev/null -I -w "%{http_code}" http://localhost:8081/streamr-core/login/auth); if [ $http_code = 200 ]; then echo "EE up and running"; break; else echo "EE not receiving connections"; sleep 5s; fi; done
     - "$TRAVIS_BUILD_DIR/streamr-docker-dev/streamr-docker-dev/bin.sh start 5"
```
hpihkala: This line should also be unnecessary, `start engine-and-editor` has already started the same set of services as `start 5`?

Actually, to be semantically correct and make the simplest statement that `streamr-client` requires all services to be running, we could replace all the other start commands with just `"$TRAVIS_BUILD_DIR/streamr-docker-dev/streamr-docker-dev/bin.sh start 5"`, and then have the wait loops that check that engine-and-editor and data-api are up?

.travis.yml
```diff
@@ -16,7 +17,15 @@ jobs:
...
-    - sleep 5m
+    - while true; do http_code=$(curl -s -o /dev/null -I -w "%{http_code}" http://localhost:8890/); if [ $http_code = 404 ]; then echo "Data-api up and running"; break; else echo "Data API not receiving connections"; sleep 5s; fi; done
+    - docker stop streamr_dev_broker
```
hpihkala: It should not be necessary to restart the broker like this. Does it not work without it?

PR #34, opened 2018-11-22
unauthenticated client

mthambipillai:


test/integration/Session.test.js
```diff
@@ -62,7 +68,20 @@ describe('Session', () => {
             .then((sessionToken) => {
                 assert(sessionToken)
             }))
-        it('should fail to get token with no authentication', (done) => clientNone.session.getSessionToken()
+        it('should get token if set with a token', () => clientSessionToken.session.getSessionToken()
```
hpihkala: Would these tests be better off as unit tests? They're not making any requests to other systems.

test/integration/Session.test.js
```diff
@@ -40,6 +41,11 @@ describe('Session', () => {
                 password: 'tester2',
             },
         })
+        clientSessionToken = createClient({
+            auth: {
+                sessionToken: 'session-token',
+            },
+        })
         clientNone = createClient({
             auth: {},
```
hpihkala: Found a bug. `new StreamrClient()` currently fails because `options.auth` is not defined. To replicate the problem just delete this line.

PR #38, opened 2019-02-11
Message id version 30

mthambipillai:


src/Subscription.js
```diff
@@ -99,41 +86,56 @@ export default class Subscription extends EventEmitter {
...
+            const to = msg.prevMsgRef
+            debug('Gap detected, requesting resend for stream %s from %o to %o', this.streamId, from, to)
+            this.emit('gap', from, to, msg.getPublisherId(), msg.messageId.msgChainId)
         } else {
-            // Normal case where prevOffset == null || lastReceivedOffset == null || prevOffset === lastReceivedOffset
-            this.lastReceivedOffset = msg.offset
-            this.callback(msg.getParsedContent(), msg)
-            if (msg.isByeMessage()) {
-                this.emit('done')
+            const messageRef = new MessageRef(msg.messageId.timestamp, msg.messageId.sequenceNumber)
```
hpihkala: There should be a `msg.getMessageRef()`

src/Subscription.js
```diff
@@ -99,41 +86,56 @@ export default class Subscription extends EventEmitter {
...
+            let res
+            if (this.lastReceivedMsgRef[key] !== undefined) {
+                res = Subscription.compareMessageRefs(messageRef, this.lastReceivedMsgRef[key])
+            }
+            if (res && (res === -1 || res === 0)) {
```
hpihkala: Could be `if (res <= 0) {` ?

test/integration/StreamrClient.test.js
```diff
@@ -99,10 +99,14 @@ describe('StreamrClient', () => {
                     })
                 })
                 client.connection.on('UnicastMessage', (msg) => {
-                    streamMessage = msg.payload
-                    assert.strictEqual(streamMessage.parsedContent.test, 'client.publish with Stream object as arg')
+                    /* eslint-disable prefer-destructuring */
```
hpihkala: Would be nicer to do as eslint wants here

test/integration/StreamrClient.test.js
```diff
@@ -132,10 +136,14 @@ describe('StreamrClient', () => {
                 })
             })
             client.connection.on('BroadcastMessage', (msg) => {
-                streamMessage = msg.payload
-                assert.strictEqual(streamMessage.parsedContent.id, id)
+                /* eslint-disable prefer-destructuring */
```
hpihkala: and here

src/Subscription.js
```diff
@@ -99,41 +86,56 @@ export default class Subscription extends EventEmitter {
...
             this.queue.push(msg)
 
-            const from = this.lastReceivedOffset + 1
-            const to = msg.previousOffset
-            debug('Gap detected, requesting resend for stream %s from %d to %d', this.streamId, from, to)
-            this.emit('gap', from, to)
-        } else if (this.lastReceivedOffset != null && msg.offset <= this.lastReceivedOffset) {
-            // Prevent double-processing of messages for any reason
-            debug('Sub %s already received message: %d, lastReceivedOffset: %d. Ignoring message.', this.id, msg.offset, this.lastReceivedOffset)
+            const from = this.lastReceivedMsgRef[key] // cannot know the first missing message so there will be a duplicate received
```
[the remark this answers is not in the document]
mthambipillai: I think it's better to do it in a separate PR so that we have more or less one PR per feature and per JIRA issue. As soon as this is merged, I could tackle it.

test/integration/StreamrClient.test.js
mthambipillai: Doing object destructuring here throws a parsing error.

test/integration/StreamrClient.test.js
mthambipillai: same as above

README.md
```diff
@@ -166,11 +166,17 @@ Name | Description
...
-resend_from_time | Resend from a specific Date (or millisecond timestamp).
-resend_to | Can be used in conjunction with `resend_from` to limit the end of the resend. By default it is the newest message.
+resend | Object defining the resend options that are specified below.
+
+#### Resend options
+
+Name | Description
+---- | -----------
+last | Resend the previous `N` messages.
+from | Resend from a specific message reference. This option must be an object that defines a `timestamp` and can optionally define a `sequenceNumber`.
```
hpihkala: Instead of this table, examples of the 3 main cases (last, from, from-to) might be more readable, due to the various interdependencies?

```
    // Resend N most recent messages
    resend: {
        last: 10,
    }

    // Resend from a specific message reference up to the newest message
    resend: {
        from: {
            timestamp: 12345,
            sequenceNumber: 0, // optional
        }
        publisher: 'publisherId', // optional
        msgChainId: 'msgChainId', // optional
    }

    // Resend a limited range of messages
    resend: {
        from: {
            timestamp: 12345,
            sequenceNumber: 0, // optional
        },
        to: {
            timestamp: 54321,
            sequenceNumber: 0, // optional
        },
        publisher: 'publisherId', // optional
        msgChainId: 'msgChainId', // optional
    }
```

src/StreamrClient.js
```diff
@@ -171,7 +173,11 @@ export default class StreamrClient extends EventEmitter {
                 })
 
             // Check pending publish requests
-            this.publisher.sendPendingPublishRequests()
+            const publishQueueCopy = this.publishQueue.slice(0)
+            this.publishQueue = []
+            publishQueueCopy.forEach((args) => {
+                this.publish(...args).catch((err) => { throw err })
```
hpihkala: `.catch((err) => { throw err })` doesn't actually do anything? Also uncaught promise reject here I guess? Should this be something like:
```
.catch((err) => {
    // log error
    // emit error event on client: this.emit('error', 'error message...')
})
```

src/StreamrClient.js
```diff
@@ -233,8 +239,48 @@ export default class StreamrClient extends EventEmitter {
         return stream ? stream.getSubscriptions() : []
     }
 
-    async publish(streamObjectOrId, data, timestamp = Date.now()) {
-        return this.publisher.publish(streamObjectOrId, data, timestamp)
+    _getMessageCreationUtil() {
+        if (!this.msgCreationUtil) {
+            this.msgCreationUtil = new MessageCreationUtil(this.options.auth, this.signer, this.getUserInfo())
```
hpihkala: Can this be initialized already in the constructor, and the function removed?

src/StreamrClient.js
```diff
@@ -233,8 +239,48 @@ export default class StreamrClient extends EventEmitter {
...
+        }
+        return this.msgCreationUtil
+    }
+
+    getMessageChainId() {
```
hpihkala: Function needed?

src/SubscribedStream.js
```diff
@@ -16,7 +16,7 @@ export default class SubscribedStream {
         }
     }
 
-    getPublishers() {
+    async getPublishers() {
```
hpihkala: unnecessary `async`?

src/Subscription.js
```diff
@@ -44,15 +43,15 @@ export default class Subscription extends EventEmitter {
 
         // Check that multiple resend options are not given
         let resendOptionCount = 0
-        if (this.options.resend_from != null) {
-            if (!(this.options.resend_from instanceof MessageRef)) {
-                throw new Error(`resend_from option needs to be a MessageRef: ${this.options.resend_from}`)
-            }
+        if (this.options.from != null) {
```
hpihkala: Looks like the counter thing can be removed and code reduced to equivalent:
```
if (this.options.from != null && this.options.last != null) {
    throw new Error(`Multiple resend options active! Please use only one: ${JSON.stringify(this.options)}`)
}
```

Might also be worth checking for invalid condition 
`this.options.from == null && this.options.to != null`.

src/Subscription.js
```diff
@@ -44,15 +43,15 @@ export default class Subscription extends EventEmitter {
 
         // Check that multiple resend options are not given
         let resendOptionCount = 0
-        if (this.options.resend_from != null) {
```
hpihkala: `options` should be renamed to `resendOptions` for clarity

test/unit/Signer.test.js
```diff
@@ -73,41 +73,43 @@ describe('Signer', () => {
         })
         it('should sign StreamMessageV30 correctly', async () => {
             const streamMessage = new StreamMessageV30(
-                [streamId, 0, timestamp, 0, null], [timestamp - 10, 0], StreamMessage.CONTENT_TYPES.JSON,
+                [streamId, 0, timestamp, 0, '', 'chain-id'], [timestamp - 10, 0], StreamMessage.CONTENT_TYPES.JSON,
```
hpihkala: If we test with `publisherId` being the empty string, the test wouldn't catch a case where the `publisherId` is accidentally ignored when concatenating the payload to be signed!

mthambipillai: Here `publisherId` can be anything because it is going to be set by the `Signer` to be `signer.address` (which is the `publisherId`). This is why on the next line the expected `payload` is using `signer.address` as `publisherId`.

PR #41, opened 2019-03-15
use ethers.js instead of web3.js

mthambipillai:


src/MessageCreationUtil.js
```diff
@@ -100,16 +98,22 @@ export default class MessageCreationUtil {
         return streamMessage
     }
 
-    static computeStreamPartition(partitionCount, partitionKey) {
+    hash(stringToHash) {
+        if (this.cachedHashes[stringToHash] === undefined) {
+            this.cachedHashes[stringToHash] = crypto.createHash('md5').update(stringToHash).digest()
```
hpihkala: Now, `cachedHashes` can grow very large over time. Maybe instead of a plain object, store them into [some simple cache that support expiration](https://www.npmjs.com/package/node-cache)?

src/Signer.js
```diff
@@ -3,30 +3,23 @@ import { MessageLayer } from 'streamr-client-protocol'
...
             this.sign = async (d) => {
-                const w3 = new Web3(this.options.provider)
-                const accounts = await w3.eth.getAccounts()
-                const address = accounts[0]
-                if (!address) {
-                    throw new Error('Cannot access account from provider')
-                }
-                this.address = address
-                return w3.eth.personal.sign(d, this.address)
+                const provider = new ethers.providers.Web3Provider(this.options.provider)
```
hpihkala: Do we need to recreate everything inside the `sign` function? Maybe `provider` and `signer` can be created once and reused?

```
const provider = new ethers.providers.Web3Provider(this.options.provider)
const signer = provider.getSigner()
this.address = signer.address
this.sign = async (d) => {
                return signer.signMessage(d)
}
```

I'm not sure how this behaves if the active address changes (for example the user selects another account on Metamask). This could be tested to get a better understanding. Maybe that changes the `signer`, in which case we could still do:

```
const provider = new ethers.providers.Web3Provider(this.options.provider)
this.sign = async (d) => {
                const signer = provider.getSigner()
                this.address = signer.address
                return signer.signMessage(d)
}
```

test/unit/Signer.test.js
```diff
@@ -6,9 +6,9 @@ const { StreamMessage, StreamMessageV30, StreamMessageV29 } = MessageLayer
 
 describe('Signer', () => {
     describe('construction', () => {
-        it('should sign when constructed with private key', () => {
+        it('should sign when constructed with private key', async () => {
             const signer = new Signer({
-                privateKey: '348ce564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709',
+                privateKey: '0x348ce564d427a3311b6536bbcff9390d69395b06ed6c486954e971d960fe8709',
```
hpihkala: Private keys passed to the client should work both with `0x` prefix and without. There hardly is a standard representation for private keys - in my experience without the prefix is more common. Please include test cases both with and without the prefix.

src/MessageCreationUtil.js
hpihkala: Ah never mind: this is in the client, so it won't see that many different values. Somehow I was thinking this was server-side.

PR #45, opened 2019-03-27
Add failing client reconnection test.

timoxley:
I believe this test should pass. Client manages to successfully connect and disconnect, but never seems to be able to connect again.


hpihkala:
@mthambipillai could you take a look at this?

PR #46, opened 2019-04-01
Add failing disconnect before connected test.

timoxley:
Perhaps this *should* fail, but currently it fails with an unhandled, uncatchable error, which is no good:

```
Unhandled "error" event. (Error: WebSocket was closed before the connection was established)
```

As a starting point, it should probably attach an error handler to `this.socket`.

### DIFF
--- a/test/integration/StreamrClient.test.js
+++ b/test/integration/StreamrClient.test.js
@@ -19,6 +19,14 @@ const createClient = (opts = {}) => new StreamrClient({
 })
 
 describe('StreamrClient Connection', () => {
+    it('can disconnect before connected', async (done) => {
+        const client = createClient()
+        client.once('error', done)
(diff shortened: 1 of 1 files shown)


hpihkala:
@timoxley @mthambipillai has this been already addressed in #52?

test/integration/StreamrClient.test.js
```diff
@@ -19,6 +19,14 @@ const createClient = (opts = {}) => new StreamrClient({
 })
 
 describe('StreamrClient Connection', () => {
+    it('can disconnect before connected', async (done) => {
+        const client = createClient()
+        client.once('error', done)
```
hpihkala: Two `done`s here now - I guess the behavior we want is that no error is thrown, so remove this line

PR #49, opened 2019-04-01
Event order tests

timoxley:
This adds failing tests for subscription event ordering.

The resent message events should probably be emitted before the final `resent` event happens, currently it seems to always send `resending`, `resent` *before* the resent messages are emitted.

Also, currently readme says `resending` will be sent before `no_resend`:
image

This doesn't seem to happen, `resending` only seems to fire before a `resent` event.

----

Additionally, I'm seeing the following 500 errors in my local tests:

```

  ● Subscription › resending/resent events › fires events in correct order

    Request to http://localhost:8081/streamr-core/api/v1/streams/KgXNomPJT9KtRF1rWVKSTA returned with error code 500: {"code":"MySQLTransactionRollbackException","message":"Deadlock found when trying to get lock; try restarting transaction"}

      34 |         return authFetch(url, session, opts, true)
      35 |     } else {
    > 36 |         throw new Error(`Request to ${url} returned with error code ${res.status}: ${text}`)
         |               ^
      37 |     }
      38 | }
      39 |
```

```
● Subscription › resending/resent events › fires events in correct order

    Request to http://localhost:8081/streamr-core/api/v1/streams/nB4vyhQtSyy5PeD-UCYPGA returned with error code 500: {"code":"StaleObjectStateException","message":"Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect): [com.unifina.domain.security.Permission#6240]"}

      34 |         return authFetch(url, session, opts, true)
      35 |     } else {
    > 36 |         throw new Error(`Request to ${url} returned with error code ${res.status}: ${text}`)
         |               ^
      37 |     }
      38 | }
      39 |

      at _callee$ (src/rest/utils.js:36:15)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:296:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
```

These don't occur if tests are run individually and seem to stem from this test teardown code:

https://github.com/streamr-dev/streamr-client-javascript/blob/f5c95d5415903b617fcd47a1bacc6a87af188ebb/test/integration/Subscription.test.js#L39-L53

### DIFF
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -76,15 +76,15 @@ export default class Subscription extends EventEmitter {
 
     // all the handleXXX methods should return promise for consistency
 
-    async handleBroadcastMessage(msg, verificationPromise) {
-        return this._handleMessage(msg, verificationPromise, false)
+    async handleBroadcastMessage(msg, verifyFn) {
+        return this._handleMessage(msg, verifyFn, false)
     }
 
-    async handleResentMessage(msg, verificationPromise) {
+    async handleResentMessage(msg, verifyFn) {
         if (!this.resending) {
             throw new Error(`There is no resend in progress, but received resent message ${msg.serialize()}`)
         } else {
-            const handleMessagePromise = this._handleMessage(msg, verificationPromise, true)
+            const handleMessagePromise = this._handleMessage(msg, verifyFn, true)
             this._lastMessageHandlerPromise = handleMessagePromise
             return handleMessagePromise
         }
@@ -106,11 +106,11 @@ export default class Subscription extends EventEmitter {
         }
 
         // Delay event emission until the last message in the resend has been handled
-        return this._lastMessageHandlerPromise.then(() => {
+        await this._lastMessageHandlerPromise.then(async () => {
             try {
                 this.emit('resent', response)
             } finally {
-                this._finishResend()
+                await this._finishResend()
             }
         })
     }
@@ -122,17 +122,17 @@ export default class Subscription extends EventEmitter {
         try {
             this.emit('no_resend', response)
         } finally {
-            this._finishResend()
+            await this._finishResend()
         }
     }
 
-    _finishResend() {
+    async _finishResend() {
         this._lastMessageHandlerPromise = null
         this.setResending(false)
-        this.checkQueue()
+        await this.checkQueue()
     }
 
-    async _handleMessage(msg, verificationPromise, isResend = false) {
+    async _handleMessage(msg, verifyFn, isResend = false) {
         if (msg.version !== 30) {
             throw new Error(`Can handle only StreamMessageV30, not version ${msg.version}`)
         }
@@ -141,7 +141,7 @@ export default class Subscription extends EventEmitter {
         }
 
         // Make sure the verification is done before proceeding
-        const valid = await verificationPromise
+        const valid = await verifyFn()
         if (!valid) {
             const err = new InvalidSignatureError(msg)
             this.handleError(err)
@@ -193,7 +193,7 @@ export default class Subscription extends EventEmitter {
         }
     }
 
-    checkQueue() {
+    async checkQueue() {
         if (this.queue.length) {
             debug('Attempting to process %d queued messages for stream %s', this.queue.length, this.streamId)
 
@@ -201,7 +201,8 @@ export default class Subscription extends EventEmitter {
             this.queue = []
 
             // Queued messages are already verified, so pass true as the verificationPromise
-            originalQueue.forEach((msg) => this._handleMessage(msg, true, false))
+            const promises = originalQueue.map((msg) => this._handleMessage(msg, () => true, false))
+            await Promise.all(promises)
         }
     }
 
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -133,7 +133,6 @@ export default class Subscription extends EventEmitter {
                 }
             })
         })
-
     }
 
     async handleNoResend(response) {
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -1,6 +1,7 @@
 import EventEmitter from 'eventemitter3'
 import debugFactory from 'debug'
 import { Errors } from 'streamr-client-protocol'
+import InvalidSignatureError from './errors/InvalidSignatureError'
 
 const debug = debugFactory('StreamrClient::Subscription')
 
@@ -57,22 +58,6 @@ export default class Subscription extends EventEmitter {
             this.setResending(false)
         })
 
-        this.on('no_resend', (response) => {
-            debug('Sub %s no_resend: %o', this.id, response)
-            this.setResending(false)
-            this.checkQueue()
-        })
-
-        this.on('resent', (response) => {
-            debug('Sub %s resent: %o', this.id, response)
-            this.setResending(false)
-            this.checkQueue()
-        })
-
-        this.on('connected', () => {
-
-        })
-
         this.on('disconnected', () => {
             this.setState(Subscription.State.unsubscribed)
             this.setResending(false)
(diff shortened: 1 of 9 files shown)


test/integration/Subscription.test.js
```diff
@@ -0,0 +1,151 @@
...
+            await client.connect()
+        })
+    })
+
+    describe('resending/no_resend events', () => {
+        it('fires events in correct order', async (done) => {
+            const subscriptionEvents = createMonitoredSubscription()
+            subscription.on('no_resend', async () => {
+                await wait(0)
+                expect(subscriptionEvents).toEqual([
```
hpihkala: `resending` and `no_resend` never appear together. It's either (`resending` AND `resent`) OR `no_resend`, depending on whether messages to resend are found or not.

hpihkala:
I looked into this enough to understand what's happening: the preprocessing of messages contains some async computations (mainly, signature verification), leading to a situation where the `resent` event occurs before any of the messages do. So this is what happens in the third test:

- `resending` message arrives and triggers event immediately
- `message1` arrives and goes into some async processing
- `message2` arrives and goes into some async processing
- `resent` message arrives and triggers event immediately
- `message1` is done with async processing and the message handler gets called
- `message2` is done with async processing and the message handler gets called

@mthambipillai since the `resent` message mutates the state of `Subscription` (for example, sets the internal `resending` flag to false), this may introduce issues given that the flag gets reset before the resent messages are processed.

hpihkala:
The essence of the fix are the changes in `Subscription.js`. Also includes some refactoring to make the code more DRY and readable.

We need to delay the `Subscription` state change and firing of the `resent` event until all the events in the resend have been processed (=the event handler has been called for them). This was a bit tricky due to the async processing happening. Proposed implementation keeps track of what is the Promise which resolves when the last message in the resend gets handled. When the `ResendResponseResent` message occurs on the websocket, we don't conclude the resend immediately like before, but rather wait until the resolution of that Promise.

@timoxley I can't add you as a reviewer since it's your PR, but please take a look at my changes and let me know if you approve.

test/integration/Subscription.test.js
timoxley: Ok, that's fine, documentation in readme needs to change though.

src/Subscription.js
```diff
@@ -89,14 +74,72 @@ export default class Subscription extends EventEmitter {
...
+        this._finishResend()
+    }
+
+    _finishResend() {
+        this._lastMessageHandlerPromise = null
+        this.setResending(false)
+        this.checkQueue()
+    }
+
+    async _handleMessage(msg, verificationPromise, isResend = false) {
```
timoxley: The way this verificationPromise is passed around is highly abnormal, makes it difficult to verify visually whether the promise gets sequenced correctly. Also looks like it can lead to unhandled rejections being reported if the verification fails and `const valid = await verificationPromise` is never called.

I think if the verification was passed as function that would make things easier to follow and much more "normal":

```js
const valid = await verify()
```
Would require minimal change to existing code & tests.

This also means verification won't even start if the handler fails for some other reason, removing the possible unhandled rejection and also avoids doing verifications that can never be used.

timoxley: Another idea is an API like this:

```js
async isVerified(msg) {
  this._verifiedMessages = this._verifiedMessages || new WeakMap()

  const stream = this.subscribedStreams[msg.streamMessage.getStreamId()]
  const isVerifiedPromise = this._verifiedMessages.get(message) || stream.verifyStreamMessage(msg.streamMessage)
  this._verifiedMessages.set(msg, isVerifiedPromise)
  return isVerifiedPromise
}
```

However the above code won't work as-is since `Subscription` doesn't currently have access to the stream in order to call `stream.verifyStreamMessage`.

src/Subscription.js
```diff
@@ -89,14 +74,72 @@ export default class Subscription extends EventEmitter {
...
+                this._finishResend()
+            })
+        } else {
+            throw new Error('Attempting to handle ResendResponseResent, but no messages have been received!')
+        }
+    }
+
+    handleNoResend(response) {
+        if (!this.resending) {
+            throw new Error(`There should be no resend in progress, but received ResendResponseNoResend message ${response.serialize()}`)
```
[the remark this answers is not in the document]
hpihkala: We don't actually need to call `_finishResend` here, because the flag shows that no resend is in progress and thus there is nothing to clean up. 

This is a sanity check, which would basically occur if the socket spontaneously sent a phantom resend response without any such request being sent. It should never occur unless something's broken either in client code or server code. 

I added these assertions because they would have caught the issue fixed in this PR earlier, and it would not have creeped in undetected as a side effect of the signature verification feature.

src/Subscription.js
hpihkala: The key goal here is that `stream.verifyStreamMessage(msg.streamMessage)` only gets called once. If we pass it wrapped into a function (`() => stream.verifyStreamMessage(msg.streamMessage)`), or pass `stream` to the handler to give `Subscription` access to it, each individual handler would verify the message separately, which we don't want.

So effectively we want to cache the result of verification, and passing in the promise which each handler can query for the result seemed like by far the simplest solution. I'm not sure why it is so unusual: it's actually very handy, although as you said it might not be very readable. We could use the `WeakMap` solution, although IMO it is much more complicated. Note that we'd need to share that `WeakMap` instance among all `Subscription` objects, it can't be an instance variable as in your example.

We could pass in a function which returns a singleton promise, which would of course be a bit silly compared to just passing in the promise. But would that make the interface more conventional?

We could also wrap the `stream.verifyStreamMessage(msg.streamMessage)` call into a more complex function that only executes `stream.verifyStreamMessage(msg.streamMessage)` on the first call, and thereafter returns a cached promise on every subsequent call, but that too would just add a function wrapping around what essentially is returning the same promise to each handler, which is what we're doing already.

hpihkala: > We could also wrap the stream.verifyStreamMessage(msg.streamMessage) call into a more complex function that only executes stream.verifyStreamMessage(msg.streamMessage) on the first call

This is what I ended up doing. The code ended up relatively clean. I was able to use a ready-made package `once`. I'm glad I discovered this package: I've been missing it for a while, and have seen people roll their own.

hpihkala:
A remaining loose end are potential unhandled rejections from the message handlers. They should probably be handled and converted to error events on the `Subscription`. I'll get back to that asap, need to leave it here for today.

src/Subscription.js
```diff
@@ -74,11 +74,13 @@ export default class Subscription extends EventEmitter {
...
 
-    handleBroadcastMessage(msg, verificationPromise) {
+    // all the handleXXX methods should return promise for consistency
+
+    async handleBroadcastMessage(msg, verificationPromise) {
         return this._handleMessage(msg, verificationPromise, false)
     }
 
-    handleResentMessage(msg, verificationPromise) {
+    async handleResentMessage(msg, verificationPromise) {
```
timoxley: 👍

src/Subscription.js
```diff
@@ -88,34 +90,40 @@ export default class Subscription extends EventEmitter {
         }
     }
 
-    handleResending(response) {
+    async handleResending(response) {
```
timoxley: 👍

test/unit/Subscription.test.js
```diff
@@ -19,6 +19,16 @@ const createMsg = (
...
 
+const assertRejects = async (promise) => {
+    let thrown
+    try {
+        await promise
+    } catch (err) {
+        thrown = err
+    }
+    assert(thrown != null)
+}
```
timoxley: Since we're using jest, you can use this form: `await expect(promise).rejects.toBeTruthy()`

src/Subscription.js
```diff
@@ -88,34 +90,40 @@ export default class Subscription extends EventEmitter {
...
         if (!this.resending) {
             throw new Error(`There should be no resend in progress, but received ResendResponseNoResend message ${response.serialize()}`)
         }
-        this.emit('no_resend', response)
-        this._finishResend()
+        try {
+            this.emit('no_resend', response)
+        } finally {
+            this._finishResend()
+        }
```
timoxley: 👍

PR #51, opened 2019-04-01
Don't suppress autoconnect on subscribe connect() errors.

timoxley:
If there's any error while connecting, the error is vanished. No way to see the error or even discover it existed without enabling the debugger and turning on "pause on caught exceptions". Not great when  to discover useful debugging information was intentionally suppressed. Just knowing that there's a pathway here for an exception to completely vanish means you need to check for this every time there's some unexpected behaviour, just in case.

I assume the specific error this was intended to suppress was the "already connecting" error. An alternative option would be to change this code to match against that specific error, but perhaps a better solution is to only attempt to connect if the client is disconnected.


src/StreamrClient.js
```diff
@@ -277,8 +277,12 @@ export default class StreamrClient extends EventEmitter {
             return this._requestPublish(streamMessage, sessionToken)
         } else if (this.options.autoConnect) {
             this.publishQueue.push([streamId, data, timestamp, partitionKey])
-            return this.connect().catch(() => {}) // ignore
+            if (this.connection.state === Connection.State.DISCONNECTED) {
```
hpihkala: This changes the logic if the connection happens to be `DISCONNECTING`. If `autoConnect` is true and we're publishing, we should probably (re-)connect in this case.

```suggestion
            if (this.connection.state === Connection.State.DISCONNECTED || this.connection.state === Connection.State.DISCONNECTING) {
```

src/StreamrClient.js
```diff
@@ -333,7 +337,9 @@ export default class StreamrClient extends EventEmitter {
         if (this.connection.state === Connection.State.CONNECTED) {
             this._resendAndSubscribe(sub)
         } else if (this.options.autoConnect) {
```
hpihkala: Code simplifies to:
```
        } else if (this.options.autoConnect && this.connection.state === Connection.State.DISCONNECTED) {
            this.connect()
        }
```

Plus adding `DISCONNECTING` case here too:

```
        } else if (this.options.autoConnect && (this.connection.state === Connection.State.DISCONNECTED || this.connection.state === Connection.State.DISCONNECTING)) {
            this.connect()
        }
```

src/StreamrClient.js
timoxley: @hpihkala Currently it seems to hang if `connect()` is called while in `DISCONNECTING` state, see https://github.com/streamr-dev/streamr-client-javascript/pull/52

timoxley: Updated in 055f55a. Rather than testing for `DISCONNECTED` or `DISCONNECTING`, changed to `!== CONNECTING`, which is equivalent since we just tested for `CONNECTED` in the `isConnected()`.

timoxley:
@hpihkala factored out the "connect if not connected, wait if connecting" conditional behaviour in `publish`/`subscribe` into a method and added some other conveniences in #56. Happy to go with this PR as-is if those changes prove controversial.

hpihkala:
@timoxley Looks like overlap with #54, where the `publish` promise resolves after publishing the messages from the publish queue, instead of resolving the `publish` promise on `connected` like is done here. For this part the approach in #54 sounds more appropriate?

timoxley:
@hpihkala yeah there's some overlap for `publish` but it's trivial and can't really be resolved until one of the PRs lands.

PR #52, opened 2019-04-02
Add failing test for trying to connect() during a disconnect().

timoxley:
Related to [this comment](https://github.com/streamr-dev/streamr-client-javascript/pull/51#discussion_r270842258) on #51.

Currently if `connect()` is called while it's in a `DISCONNECTING` state, the new connect never completes.

e.g. https://github.com/streamr-dev/streamr-client-javascript/blob/9f9d86a4f2024fb790397070ee9c186b430d87cf/test/integration/StreamrClient.test.js#L57-L68

The above code never gets past the `await client.connect()` inside `.on('disconnecting')`.

The sequence of state transitions for the connection goes like:

1. `CONNECTING`
2. `CONNECTED`
3. `DISCONNECTING`
4. `CONNECTING`
5. `DISCONNECTED`
6. `CONNECTING`

The final `CONNECTING` is caused by this code, not clear yet whether this is also the cause of the inability to reconnect:

https://github.com/streamr-dev/streamr-client-javascript/blob/89a261d38cc46f18c09e10e3f10bf316059dd76c/src/Connection.js#L50-L59

So IIUC the scenario goes something like this:

1. `client.disconnect()` is called (due to subscriptions empty or an explicit call)
2. While in `DISCONNECTING` state `client.connect()` is called e.g. `client.subscribe()` was called again
3. State now moves from `DISCONNECTING` to `CONNECTING`. This replaces the websocket but doesn't unsubscribe the old socket listeners.
4. Original websocket can possibly close before the new connect completes. The original socket `close` event handler (above) fires.
5. State is not `DISCONNECTING`, it's `CONNECTING`, so it (mis)reports to `debug` "connection lost, reconnecting", sets the connection state from `CONNECTING` to `DISCONNECTED` and queues a `connect()` call for 2 seconds from now.
6. The original connect (should) complete, now moving the state directly from `DISCONNECTED` to `CONNECTED` 🤷‍♀️ (though currently this `CONNECTED` never happens)
7. And to top it off, `connect()` gets called again in two seconds which will forcibly try reconnect the client even if the user had explicitly called `disconnect()` or it had itself auto-disconnected due to no subscriptions. Also possible this could be the `connect()` call that triggers this whole sequence (again) at step 2.

### DIFF
--- a/src/Connection.js
+++ b/src/Connection.js
@@ -32,6 +32,14 @@ class Connection extends EventEmitter {
         } else if (this.state === Connection.State.CONNECTED) {
             return Promise.reject(new Error('Already connected!'))
         }
+        if (this.state === Connection.State.DISCONNECTING) {
+            return new Promise((resolve) => {
+                this.once('disconnected', async () => {
+                    await this.connect()
+                    resolve()
+                })
+            })
+        }
         if (!this.socket || this.socket.readyState === WebSocket.CLOSED) {
             this.socket = new WebSocket(this.options.url)
         }
@@ -81,6 +89,14 @@ class Connection extends EventEmitter {
             return Promise.reject(new Error('Already disconnected!'))
         } else if (this.socket === undefined) {
             return Promise.reject(new Error('Something is wrong: socket is undefined!'))
+        }
+        if (this.state === Connection.State.CONNECTING) {
+            return new Promise((resolve) => {
+                this.once('connected', async () => {
+                    await this.disconnect()
+                    resolve()
+                })
+            })
         }
 
         return new Promise((resolve) => {
--- a/src/Connection.js
+++ b/src/Connection.js
@@ -34,10 +34,7 @@ class Connection extends EventEmitter {
         }
         if (this.state === Connection.State.DISCONNECTING) {
             return new Promise((resolve) => {
-                this.once('disconnected', async () => {
-                    await this.connect()
-                    resolve()
-                })
+                this.once('disconnected', () => resolve(this.connect()))
             })
         }
         if (!this.socket || this.socket.readyState === WebSocket.CLOSED) {
@@ -92,10 +89,7 @@ class Connection extends EventEmitter {
         }
         if (this.state === Connection.State.CONNECTING) {
             return new Promise((resolve) => {
-                this.once('connected', async () => {
-                    await this.disconnect()
-                    resolve()
-                })
+                this.once('connected', () => resolve(this.disconnect()))
             })
         }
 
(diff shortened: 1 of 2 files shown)


timoxley:
For an example that doesn't call connect/disconnect directly, instead uses autoconnect: 

```js
describe('autoconnect + subscriptions', () => {
    it('can reconnect if subscription added while disconnecting', async (done) => {
        const client = createClient({
            autoConnect: true,
            autoDisconnect: true,
        })

        client.once('connected', async () => {
            const stream = await client.createStream({
                name: uniqueId(),
            })
            const subscription = client.subscribe({
                stream: stream.id,
                resend: RESEND_ALL,
            }, () => {})

            subscription.once('no_resend', async () => {
                client.once('disconnecting', async () => {
                    // client disconnecting after unsubscribing
                    client.once('connected', async () => {
                        // client re-connected after subscribing
                        await client.disconnect()
                        done()
                    })
                    client.subscribe({
                        stream: stream.id,
                        resend: RESEND_ALL,
                    }, () => {})
                })
                await client.unsubscribe(subscription)
            })
        })
    })
})
```

src/Connection.js
```diff
@@ -32,6 +32,14 @@ class Connection extends EventEmitter {
         } else if (this.state === Connection.State.CONNECTED) {
             return Promise.reject(new Error('Already connected!'))
         }
+        if (this.state === Connection.State.DISCONNECTING) {
+            return new Promise((resolve) => {
+                this.once('disconnected', async () => {
+                    await this.connect()
```
timoxley: if connect rejects, then I think this will deadlock. I think it should be: `resolve(this.connect())`. If connect rejects, then this will resolve to a rejection which is equivalent to calling `reject`

src/Connection.js
```diff
@@ -82,6 +90,14 @@ class Connection extends EventEmitter {
         } else if (this.socket === undefined) {
             return Promise.reject(new Error('Something is wrong: socket is undefined!'))
         }
+        if (this.state === Connection.State.CONNECTING) {
+            return new Promise((resolve) => {
+                this.once('connected', async () => {
+                    await this.disconnect()
```
timoxley: same here

test/integration/StreamrClient.test.js
```diff
@@ -119,7 +119,7 @@ describe('StreamrClient Connection', () => {
                 // wait for possible reconnections
                 setTimeout(() => {
                     client.off('connecting', onConnecting)
-                    expect(client.isConnected()).toBeFalse()
+                    expect(client.isConnected()).toBe(false)
```
timoxley: 👍

src/Connection.js
timoxley: also what should happen if there's an 'error' event before 'disconnected' happens? It should probably remove the once('disconnected' handler on error. Event emitters are fraught with peril.

hpihkala:
@timoxley happy with changes? Let's get this merged and revisit #51 which is related

PR #53, opened 2019-04-02
fetch and cache streams in MessageCreationUtil

mthambipillai:


package.json
```diff
@@ -63,6 +63,7 @@
     "debug": "^3.1.0",
     "ethers": "^4.0.27",
     "eventemitter3": "^3.0.1",
+    "node-cache": "^4.2.0",
```
timoxley: Note `node-cache` depends on `lodash`, would be interested how much this affects the browser bundle size.

mthambipillai: 821KB on `master` --> 844KB on this branch

src/MessageCreationUtil.js
```diff
@@ -22,6 +30,23 @@ export default class MessageCreationUtil {
...
+    async getStream(streamId) {
+        if (!this.cachedStreams.get(streamId)) {
+            if (!this.ongoingRequests[streamId]) {
+                this.ongoingRequests[streamId] = this.getStreamFunction(streamId)
+            }
+            const stream = await this.ongoingRequests[streamId]
+            const success = this.cachedStreams.set(streamId, {
+                id: stream.id,
+                partitions: stream.partitions,
+            })
```
timoxley: So `cachedStreams` is supposed to evict the stream after 30 minutes, presumably to prevent memory leaks, but the stream can never be garbage collected because a reference will still be held in the `this.ongoingRequests[streamId]` promise. 

As it stands, you could actually just remove `node-cache` because `ongoingRequests` is already acting like a cache for you, only thing it's not doing is eviction.

Anyway, it should probably do something like:
```js
let stream
try {
   stream = await this.ongoingRequests[streamId]
} finally {
  delete this.ongoingRequests[streamId]
}
```

src/MessageCreationUtil.js
```diff
@@ -22,6 +30,23 @@ export default class MessageCreationUtil {
...
+            if (!success) {
+                throw new Error(`Could not store stream with id ${streamId} in local cache.`)
```
timoxley: Does it need to crash here? it could just log a warning and return the stream.

package.json
timoxley: Oh man ok it's huge already

src/MessageCreationUtil.js
```diff
@@ -32,16 +31,14 @@ export default class MessageCreationUtil {
...
-            if (!this.ongoingRequests[streamId]) {
-                this.ongoingRequests[streamId] = this.getStreamFunction(streamId)
-            }
-            const stream = await this.ongoingRequests[streamId]
-            const success = this.cachedStreams.set(streamId, {
+            const streamPromise = this.getStreamFunction(streamId).then((stream) => ({
                 id: stream.id,
                 partitions: stream.partitions,
-            })
+            }))
```
timoxley: Nice one, immediately cache the promise not the stream and avoid any timing issues. Nice.

timoxley: Does `node-cache`'s `useClones: true` affect this in any way?

src/MessageCreationUtil.js
```diff
@@ -1,15 +1,22 @@
...
 import randomstring from 'randomstring'
 import { MessageLayer } from 'streamr-client-protocol'
 import { ethers } from 'ethers'
+import Stream from './rest/domain/Stream'
 
 const { StreamMessage } = MessageLayer
 
 export default class MessageCreationUtil {
-    constructor(auth, signer, userInfoPromise) {
+    constructor(auth, signer, userInfoPromise, getStreamFunction) {
```
timoxley: 4 arguments is usually where I draw the line and  switch to named arguments i.e. options object, might be too much to change though.

src/MessageCreationUtil.js
mthambipillai: I guess it wouldn't hurt to set it to `false` since we don't modify the values after retrieving them from the cache. But probably safer to keep it to `true` unless we need cache a complex mutable object, the performance cost of cloning should be very small for this object (only 2 fields)

timoxley: the promise is what's stored in the cache so it's "cloning" the promise (whatever that means), not the object with two fields, right?

mthambipillai: Yes the promise actually

hpihkala: Can cloning promises even work? When the original promise is resolved, is the clone resolved too? I feel slightly uneasy just thinking of a cache returning cloned promises on `get`.

hpihkala:
@timoxley happy with Melchior's changes as per your feedback?

PR #54, opened 2019-04-03
StreamrClient#publish: queue size limit & resolving when published

harbu:
Two changes in two separate commits regarding `StreamrClient#publish`. Please give feedback.

1. If, for whatever reason, streamr-client is unable to (re-)establish connection to Data API then `maxPublishQueueSize` prevents queue from growing without bounds.

2. If invoking `publish` while streamr-client is disconnected from Data API wait for publishing to occur before resolving Promise.

### DIFF
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -189,14 +189,7 @@ export default class StreamrClient extends EventEmitter {
             // Check pending publish requests
             const publishQueueCopy = this.publishQueue.slice(0)
             this.publishQueue = []
-            publishQueueCopy.forEach(([streamId, data, timestamp, partitionKey, resolve, reject]) => {
-                this.publish(streamId, data, timestamp, partitionKey).then(resolve)
-                    .catch((err) => {
-                        debug(`Error: ${err}`)
-                        this.emit(err)
-                        reject(err)
-                    })
-            })
+            publishQueueCopy.forEach((publishFn) => publishFn())
         })
 
         this.connection.on('disconnected', () => {
@@ -282,8 +275,16 @@ export default class StreamrClient extends EventEmitter {
                     `publishQueue exceeded maxPublishQueueSize=${this.options.maxPublishQueueSize}`,
                 )
             }
-            this.publishQueue.push([streamId, data, timestamp, partitionKey])
-            return this.ensureConnected()
+            return new Promise((resolve, reject) => {
+                this.publishQueue.push(async () => (
+                    this.publish(streamId, data, timestamp, partitionKey)
+                        .then(resolve, (err) => {
+                            debug(`Error: ${err}`)
+                            this.emit('error', err)
+                            reject(err)
+                        })
+                ))
+            })
         }
 
         throw new FailedToPublishError(
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -189,11 +189,13 @@ export default class StreamrClient extends EventEmitter {
             // Check pending publish requests
             const publishQueueCopy = this.publishQueue.slice(0)
             this.publishQueue = []
-            publishQueueCopy.forEach((args) => {
-                this.publish(...args).catch((err) => {
-                    debug(`Error: ${err}`)
-                    this.emit(err)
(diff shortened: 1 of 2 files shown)


src/StreamrClient.js
```diff
@@ -191,11 +192,13 @@ export default class StreamrClient extends EventEmitter {
...
-            publishQueueCopy.forEach((args) => {
-                this.publish(...args).catch((err) => {
-                    debug(`Error: ${err}`)
-                    this.emit(err)
-                })
+            publishQueueCopy.forEach(([streamId, data, timestamp, partitionKey, resolve, reject]) => {
+                this.publish(streamId, data, timestamp, partitionKey).then(resolve)
+                    .catch((err) => {
+                        debug(`Error: ${err}`)
+                        this.emit(err)
```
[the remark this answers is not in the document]
harbu: Good point. From the code looks like its an error object, so it should probably be as you say.

harbu:
I agree with your proposed changes. I'll wait for more opinions though before implementing them.

timoxley:
The CI failure is interesting:

```patch
-    "publisherId": "16f78a7d6317f102bbd95fc9a4f3ff2e3249287690b8bdad6b7810f82b34ace3",
+    "publisherId": "0x16f78a7d6317f102bbd95fc9a4f3ff2e3249287690b8bdad6b7810f82b34ace3",
```

timoxley:
Is this problem shared across all commands issued while a connection is pending?

hpihkala:
> Is this problem shared across all commands issued while a connection is pending?

@timoxley The commands that need (and trigger, if `autoConnect` is true) the websocket connection are basically only `publish` and `subscribe`. (Resends of course, but they happen after `subscribed`). 

I see what you're after: could there be a unified command queue which gets executed on `connected`?

Looking at the code, currently there are separate mechanisms for handling pending publishes and pending subscribes, mainly because subscribes are one-offs and more complicated in terms of what internal bookkeeping is required. What currently happens with subscriptions looks like a pretty reasonable approach: the existing subscription objects are examined on `connected`, and the appropriate subscribe commands are sent to the server - the subscriptions are kind of synced with the server. Multiple local subscriptions are multiplexed over the single subscribe command to the server, so there is not a one-to-one correspondence between `subscribe` method calls and subscribe messages to be sent to the server, like there is with `publish`.

While a worthy idea to try and generalize how pending commands and state is handled, I wouldn't dive into refactoring it at this point, as the handling of subscriptions require more complicated logic than simply queuing commands.

src/StreamrClient.js
```diff
@@ -276,8 +272,23 @@ export default class StreamrClient extends EventEmitter {
...
         } else if (this.options.autoConnect) {
-            this.publishQueue.push([streamId, data, timestamp, partitionKey])
-            return this.connect().catch(() => {}) // ignore
+            if (this.publishQueue.length >= this.options.maxPublishQueueSize) {
+                throw new FailedToPublishError(
+                    streamId,
+                    data,
+                    `publishQueue exceeded maxPublishQueueSize=${this.options.maxPublishQueueSize}`,
+                )
+            }
```
timoxley: Optional, but I feel it would be polite to provide some API that could be used to check before the user calls publish? Otherwise it seems like the only way to work around this is to wait for this exception, hope you detected this specific exception correctly and respond from there i.e. exception based flow control.

This is to catch edge cases and ideally users would raise their max queue size appropriately but it's also possible they aren't in control of how many messages are being published. I guess in those cases they could set the queue size to `Infinity`.

src/StreamrClient.js
```diff
@@ -276,8 +272,23 @@ export default class StreamrClient extends EventEmitter {
...
+            return new Promise((resolve, reject) => {
+                this.publishQueue.push(async () => (
+                    this.publish(streamId, data, timestamp, partitionKey)
+                        .then(resolve, (err) => {
+                            debug(`Error: ${err}`)
+                            this.emit('error', err)
+                            reject(err)
+                        })
```
timoxley: :+1:

PR #57, opened 2019-04-05
add logout functionality

mthambipillai:


src/Session.js
```diff
@@ -47,6 +47,12 @@ export default class Session {
         }
         return this.sessionTokenPromise
     }
+
+    async logout() {
+        await this._client.logoutEndpoint()
```
timoxley: Sorry to do this to you again, but…

What happens to commands *while* it's logging out? 

e.g. 

* What happens if `logout()` is called again while it's logging out?
* What if `getSessionToken()` is called (directly or indirectly) while it's logging out?
* What if `logout()` is called before the session has finished logging in?
* Should the session token be set to `undefined` and the state to `LOGGED_OUT` if `logoutEndpoint()` fails? (currently if `logoutEndpoint` fails the following two lines won't execute)

Perhaps it should do the await step *after* unsetting the session token and setting the state.

PR #58, opened 2019-04-09
Core 1671 retry resends

mthambipillai:
- Periodic gap filling request until gap filled.
- Second resend request once subscribed if no answer (UnicastMessage or ResendResponseNoResend) received after some delay.


src/StreamrClient.js
```diff
@@ -413,6 +415,9 @@ export default class StreamrClient extends EventEmitter {
             sub.once('subscribed', () => {
                 if (sub.hasResendOptions()) {
                     this._requestResend(sub)
+                    const secondResend = setTimeout(() => this._requestResend(sub), STORAGE_DELAY)
```
timoxley: Could probably do with a comment or two to explain what this is for e.g. adapt from jira ticket

src/Subscription.js
```diff
@@ -56,13 +60,25 @@ export default class Subscription extends EventEmitter {
...
+            Object.keys(this.gaps).forEach((gap) => {
+                clearInterval(gap)
+            })
             this.setResending(false)
         })
+
+        this.on('error', () => {
+            Object.keys(this.gaps).forEach((gap) => {
+                clearInterval(gap)
+            })
```
timoxley: This piece of code is repeated a lot and should probably be factored out e.g. `stopAllGapPolling()` or something. You should also `delete this.gaps[gap]` otherwise you got a mem leak.

src/StreamrClient.js
```diff
@@ -413,6 +415,9 @@ export default class StreamrClient extends EventEmitter {
...
+                    sub.once('message received', () => clearTimeout(secondResend))
+                    sub.once('no_resend', () => clearTimeout(secondResend))
```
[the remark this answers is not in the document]
timoxley: > it really should clean up any handlers that can no longer be used, otherwise the subscription can't be garbage collected.

TODO

src/Subscription.js
```diff
@@ -196,6 +214,18 @@ export default class Subscription extends EventEmitter {
             }
             debug('Gap detected, requesting resend for stream %s from %o to %o', this.streamId, from, to)
             this.emit('gap', fromObject, toObject, msg.getPublisherId(), msg.messageId.msgChainId)
+
+            // If for some reason the missing messages are not received, the gap filling request is resent every 'resendTimeout' seconds
+            // until a message is received, at which point the gap will be filled or
+            // a new different gap request will be sent and resent every 'resendTimeout' seconds.
+            clearInterval(this.gaps[key])
+            this.gaps[key] = setInterval(() => {
+                if (this.lastReceivedMsgRef[key].compareTo(to) === -1) {
```
timoxley: Not related to this PR but I do note that items are never removed from `this.lastReceivedMsgRef` so it'll only ever grow in size until the subscription can be garbage collected.

mthambipillai: Normally we need to remember the reference of the last received message per publisher and message chain id for gap filling / duplicate detection / ordering, so we cannot really delete them. If a publisher is removed, we could delete all related entries (periodically fetch the list of publishers of the stream).
But for the message chain id, we cannot really know if a publisher is ever going to use the same again (unless we receive a "bye" message). A publisher could publish with this chain id only once per day, so if we use a standard expiration time for every key it should be very large. We could also compute the mean `m` and standard deviation `d` of the publication period for each chain id and determine some key-specific expiration time using a linear combination of `m` and `d` like in TCP timeout based on RTT.
A buggy/malicious publisher could also flood messages with a new chain id every time, so there could be a counter for a max number of chain ids per publisher and refuse new chain ids.
Would be good to have @hpihkala and @harbu opinions on this matter and address it in some other PR if this is really an issue.

timoxley: Yep not something for this pr

hpihkala: In practice we can assume that the set of `(publisher, msgChainId)` tuples has an upper bound, and client session times are not infinite.

However this does actually open an attack where an evil publisher could rotate their `msgChainId` on every message in an attempt to exhaust memory on subscribed clients. Clients could of course defend by stopping to listen to such streams, so the attack would be kind of feeble.

I don't feel that this needs to be addressed right now in this PR, but @mthambipillai please log a JIRA ticket with minor priority for rethinking whether we could do some kind of evictions from the last received message cache. The evictions can't be based on time for the reasons Melchior mentioned above, but one idea: we could have a hard limit on the number of remembered `msgChainId`s per publisher, which could be set relatively high (e.g. 1000) to defend against the above mentioned attack/publisher malfunction, while allowing for almost all valid cases.

mthambipillai: Yes the attack you describe and the hard limit solution is what I meant by
> A buggy/malicious publisher could also flood messages with a new chain id every time, so there could be a counter for a max number of chain ids per publisher and refuse new chain ids.

hpihkala: Yes, haha, looks like my subconscious read your message more closely than my conscious mind :)

harbu: 👍 for hard limit

src/StreamrClient.js
```diff
@@ -485,10 +485,21 @@ export default class StreamrClient extends EventEmitter {
...
-                    sub.once('message received', () => clearTimeout(secondResend))
-                    sub.once('no_resend', () => clearTimeout(secondResend))
-                    sub.once('unsubscribed', () => clearTimeout(secondResend))
-                    sub.once('error', () => clearTimeout(secondResend))
+                    const callback = () => clearTimeout(secondResend)
+                    const removeCallbacks = () => {
+                        sub.removeListener('message received', callback)
+                        sub.removeListener('no_resend', callback)
+                        sub.removeListener('unsubscribed', callback)
+                        sub.removeListener('error', callback)
```
timoxley: Should these be removing `handler` instead of `callback`? e.g. `sub.removeListener('error', handler)`

src/Subscription.js
```diff
@@ -60,24 +60,25 @@ export default class Subscription extends EventEmitter {
...
-                clearInterval(gap)
-            })
+            this._clearGaps()
+        })
+    }
+
+    _clearGaps() {
+        Object.keys(this.gaps).forEach((gap) => {
+            clearInterval(gap)
+            delete this.gaps[gap]
```
timoxley: 👍

src/StreamrClient.js
```diff
@@ -47,6 +47,8 @@ export default class StreamrClient extends EventEmitter {
             auth: {},
             publishWithSignature: 'auto',
             verifySignatures: 'auto',
+            storageDelay: 5000,
+            resendTimeout: 5000,
```
hpihkala: Would `gapfillTimeout` be a clearer name?

src/StreamrClient.js
```diff
@@ -476,9 +478,28 @@ export default class StreamrClient extends EventEmitter {
...
-            sub.once('subscribed', () => {
+            sub.once('subscribed', async () => {
                 if (sub.hasResendOptions()) {
-                    this._requestResend(sub)
+                    await this._requestResend(sub)
+                    // the requested messages might not have been stored yet, so retry if no answer after 'storageDelay' seconds
+                    const secondResend = setTimeout(() => this._requestResend(sub), this.options.storageDelay)
+                    // once a message is received, gap filling in Subscription.js will check if this satisfies the resend and request
+                    // another resend if it doesn't. So we can anyway clear this resend request.
+                    const callback = () => clearTimeout(secondResend)
```
hpihkala: Used only once, `clearTimeout(secondResend)` could just be inline on line 490?

src/StreamrClient.js
```diff
@@ -476,9 +478,28 @@ export default class StreamrClient extends EventEmitter {
...
+                    const handler = () => {
+                        callback()
+                        const removeCallbacks = () => {
```
hpihkala: Used only once, the `removeListener` calls could just be inline?

src/StreamrClient.js
```diff
@@ -476,9 +478,28 @@ export default class StreamrClient extends EventEmitter {
...
+                            sub.removeListener('message received', handler)
+                            sub.removeListener('no_resend', handler)
+                            sub.removeListener('unsubscribed', handler)
+                            sub.removeListener('error', handler)
+                        }
+                        removeCallbacks()
+                    }
+                    sub.once('message received', handler)
+                    sub.once('no_resend', handler)
```
hpihkala: `no_resend` shouldn't actually cancel the retry! For example, let's assume there's 1 message in the stream and it's still in "storage delay":

- resend last 1
- `no_resend` (message not written to storage yet)
- cancel retry

The client wouldn't ever get the message. Instead if it retries:

- resend last 1
- `no_resend` (message not written to storage yet)
- wait for `storageDelay`
- resend last 1
- now the message is found and delivered

src/StreamrClient.js
```diff
@@ -47,6 +47,8 @@ export default class StreamrClient extends EventEmitter {
             auth: {},
             publishWithSignature: 'auto',
             verifySignatures: 'auto',
+            storageDelay: 5000,
```
hpihkala: Would `retryResendAfter` be a clearer name?

src/Subscription.js
```diff
@@ -56,13 +60,26 @@ export default class Subscription extends EventEmitter {
...
         })
+
+        this.on('error', () => {
+            this._clearGaps()
+        })
+    }
+
+    _clearGaps() {
+        Object.keys(this.gaps).forEach((gap) => {
+            clearInterval(gap)
```
hpihkala: Should this be `clearInterval(this.gaps[gap])`? (might avoid bugs and be more consistent if you use the variable name `key` instead of `gap`)

test/unit/StreamrClient.test.js
```diff
@@ -215,9 +218,12 @@ describe('StreamrClient', () => {
                     })
                     connection.expect(ResendLastRequest.create(sub.streamId, sub.streamPartition, sub.id, nbToResend))
                     connection.emitMessage(SubscribeResponse.create(sub.streamId))
+                    // sending second resend last request since no answer to the first request was received
+                    connection.expect(ResendLastRequest.create(sub.streamId, sub.streamPartition, sub.id, nbToResend))
+                    setTimeout(done, STORAGE_DELAY + 1000)
```
hpihkala: In unit tests you should never use "real" time, but instead use [fake timers](https://sinonjs.org/releases/v1.17.6/fake-timers/) so that you have full control over time and can explicitly tick it forward (and the test will run fast).

hpihkala: Same elsewhere in this file

test/unit/Subscription.test.js
```diff
@@ -241,6 +243,43 @@ describe('Subscription', () => {
...
+                        assert.deepStrictEqual(from, from2)
+                        assert.deepStrictEqual(to, to2)
+                        assert.deepStrictEqual(publisherId, publisherId2)
+                        done()
+                    })
+                })
+
+                sub.handleBroadcastMessage(msg1, sinon.stub().resolves(true))
+                sub.handleBroadcastMessage(msg4, sinon.stub().resolves(true))
+            }, RESEND_TIMEOUT + 2000)
```
hpihkala: Should use fake timers in this file too

src/Subscription.js
hpihkala: Plus add test that catches this bug (checks that the resend is not retried if the sub is disconnected or unsubscribed)

test/unit/StreamrClient.test.js
mthambipillai: I managed to use fake timers in `Subscription.test.js` but in this file it was too complex. The `tick()` function must be called after the `setTimeout()` in `StreamrClient._resendAndSubscribe()` but there is no event emitted to to catch when that happens...

PR #59, opened 2019-04-12
Handle disconnecting during publish.

timoxley:
You may have noticed the following error being logged when running the integration tests:

> `Error: WebSocket is not open: readyState 2 (CLOSING)`

image

https://travis-ci.com/streamr-dev/streamr-client-javascript/jobs/191090350#L2869-L2884

The above gets logged during running the current integration tests, yet the test suite does not fail because no test is actually listening for the client 'error' event at the time it occurs. It seems to be caused by the test suite itself not waiting for the publish action to finish before disconnecting the client. It appears the message does fail to be published.

Note the code below from the current `master` branch:

https://github.com/streamr-dev/streamr-client-javascript/blob/39db82ddcdf7e0da8917b85768a359c338704965/src/StreamrClient.js#L269-L272

`createStreamMessage` is async, so it's entirely possible that the client will fully or partially disconnect between the first `isConnected` check and `createStreamMessage` resolving. This causes the `_requestPublish` to try writing to the disconnecting websocket, which in turn triggers the above error event, which isn't caught by anything.

This PR makes `publish` re-check the client is connected try to reconnect if `autoConnect` is set, or fail with an error "Disconnected before publish" otherwise. 

Also adjusted the tests that initially exhibited this error to `await` their operations before continuing.

----

**Update**: Noticed similar behaviour for subscriptions: if you call `client.disconnect()` on `subscribed`, and the subscription had resend options, the client will cough up a similar error event as it tries to write to a closing websocket: 

```
  console.error src/StreamrClient.js:229
    Error: WebSocket is not open: readyState 2 (CLOSING)
        at WebSocket.send (/Users/timoxley/Projects/streamr/streamr-client/node_modules/ws/lib/websocket.js:322:19)
        at Connection.send (/Users/timoxley/Projects/streamr/streamr-client/src/Connection.js:108:25)
        at send (/Users/timoxley/Projects/streamr/streamr-client/src/StreamrClient.js:559:29)
        at <anonymous>
        at process._tickCallback (internal/process/next_tick.js:189:7)
```

Solved this in 1fbd71167a44d60085e529ca43f8a7ef0312f891 by simply ignoring resend requests if the client is no longer connected. A more general solution should probably be found.

### DIFF
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -254,7 +254,6 @@ export default class StreamrClient extends EventEmitter {
     }
 
     async publish(streamObjectOrId, data, timestamp = Date.now(), partitionKey = null) {
-        const sessionToken = await this.session.getSessionToken()
         // Validate streamObjectOrId
         let streamId
         if (streamObjectOrId instanceof Stream) {
@@ -265,21 +264,13 @@ export default class StreamrClient extends EventEmitter {
             throw new Error(`First argument must be a Stream object or the stream id! Was: ${streamObjectOrId}`)
         }
 
-        // If connected, emit a publish request
+        const [sessionToken, streamMessage] = await Promise.all([
+            this.session.getSessionToken(),
+            this.msgCreationUtil.createStreamMessage(streamObjectOrId, data, timestamp, partitionKey),
+        ])
+
         if (this.isConnected()) {
-            const streamMessage = await this.msgCreationUtil.createStreamMessage(streamObjectOrId, data, timestamp, partitionKey)
-            // if happened to disconnect before completed
-            if (!this.isConnected()) {
-                if (this.options.autoConnect) {
-                    // try reconnect if autoConnect
-                    return this.publish(streamId, data, timestamp, partitionKey)
-                }
-                throw new FailedToPublishError(
-                    streamId,
-                    data,
-                    'Disconnected before publish completed',
-                )
-            }
+            // If connected, emit a publish request
             return this._requestPublish(streamMessage, sessionToken)
         } else if (this.options.autoConnect) {
             if (this.publishQueue.length >= this.options.maxPublishQueueSize) {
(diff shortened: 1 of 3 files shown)


src/StreamrClient.js
```diff
@@ -268,6 +268,18 @@ export default class StreamrClient extends EventEmitter {
         // If connected, emit a publish request
         if (this.isConnected()) {
             const streamMessage = await this.msgCreationUtil.createStreamMessage(streamObjectOrId, data, timestamp, partitionKey)
+            // if happened to disconnect before completed
```
hpihkala: Ooh this is getting messy! Would be great to not to have to repeat any part of the "if not connected, then queue the message and autoconnect" logic. There must be a cleaner way. How about along the lines of:

- Create the `streamMessage` **before** the first `this.isConnected()` check on line 269
- Then if we end up in the autoconnect procedure, push a function that calls `this._requestPublish(streamMessage, sessionToken)` into the publish queue instead of the function that calls `this.publish(streamId, data, timestamp, partitionKey)`

timoxley: Done in e2e9a19

@hpihkala Note one side effect of this change is that the timestamp will now match the point where `publish()` was called, rather than when the connection happened. This is probably desirable anyway. cc @harbu

test/unit/StreamrClient.test.js
```diff
@@ -806,6 +806,7 @@ describe('StreamrClient', () => {
         })
 
         it('rejects the promise if autoConnect is false and the client is not connected', (done) => {
+            client.options.auth.username = 'username'
```
timoxley: The new publish sequence now tries to create message before checking for valid connection (i.e. valid credentials), so this test would fail on invalid auth config before it even gets a chance to fail on the connection status.

src/StreamrClient.js
harbu: Sound reasonable imo.

test/unit/StreamrClient.test.js
hpihkala: Just a note, that connecting (the websocket) does not trigger authentication sequence (`getSessionToken` does). Establishing a connection with invalid/no credentials is/should be possible.

src/StreamrClient.js
hpihkala: A lot cleaner now, thanks!

PR #60, opened 2019-04-18
Auth exceptions incl. more details about responses

mondoreale:
In this PR I add response `body` and `response` itself to the exceptions thrown by `authFetch`. Canned failure messages are not enough for Platform's login page.

I also moved `authFetch` out from `utils.js` (which is now gone) and gave it its own cozy file. `import { authFetch } from '…'` you do `import authFetch from '…'` now.

Guys, let me know what you think.


src/rest/authFetch.js
```diff
@@ -0,0 +1,35 @@
...
+        },
+    })
+
+    const body = await response.text()
+
+    if (response.ok) {
+        try {
+            return JSON.parse(body || '{}')
+        } catch (e) {
+            throw new AuthFetchError(`Failed to parse JSON response: ${body}`, response, body)
```
timoxley: This is inherited from the existing code but consider the case when `body` is a massive chunk of html, the error message will be pretty unusable. Perhaps should just leave it out of the message entirely since we're attaching to the error object, or limit the damage with `body.slice(0, 100)`.

src/rest/authFetch.js
```diff
@@ -0,0 +1,35 @@
...
+        }
+    } else if ([400, 401].includes(response.status) && !requireNewToken) {
+        return authFetch(url, session, opts, true)
+    } else {
+        throw new AuthFetchError(`Request to ${url} returned with error code ${response.status}: ${body}`, response, body)
```
timoxley: Same here

src/errors/AuthFetchError.js
```diff
@@ -0,0 +1,9 @@
+class AuthFetchError extends Error {
+    constructor(message, response, body) {
+        super(message)
+        this.response = response
+        this.body = body
```
timoxley: Note this may not capture the stack trace correctly: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types

src/rest/authFetch.js
timoxley: Also possible this could re-use the existing JSON parse message `throw new AuthFetchError(e.message, response, body)`

src/errors/AuthFetchError.js
```diff
@@ -0,0 +1,9 @@
...
+    }
+}
+
+module.exports = AuthFetchError
```
timoxley: module.exports?

timoxley:
@mondoreale @hpihkala bump

hpihkala:
Tim's review is enough, I don't think you need me or Melchior for this. Please proceed.

src/errors/AuthFetchError.js
mondoreale: Using `captureStackTrace` now. It's not part of `babel-plugin-transform-builtin-extend` out of the box. Should fly now. 🚀

src/errors/AuthFetchError.js
mondoreale: Wanted to make it similar to other errors. Changed to `export default` now. It'd be good to be consistent here tho. It's a good material for a future refactoring.

PR #62, opened 2019-05-06
CORE-1707: Fix unauthenticated mode

harbu:
Fix bug in which streamr-client asks E&E for userInfo, i.e. `GET https://www.streamr.com/api/v1/users/me`, even if user has provided no api key.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -29,6 +29,10 @@ export default class Session extends EventEmitter {
         }
     }
 
+    isUnauthenticated() {
+        return this.options.unauthenticated
+    }
+
     updateState(newState) {
         this.state = newState
         this.emit(newState)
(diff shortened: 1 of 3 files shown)


harbu:
Thank you @mthambipillai for fixing tests! 👍

harbu:
@hpihkala Not sure if your comment was referring to fixing the existing tests or writing new ones. If the latter, I wasn't planning on writing new ones for this change. Is there some specific scenario you think should be tested here?

harbu:
I'm gonna land this now because this is blocking my work on broker.

hpihkala:
Sorry, what I meant was that a test should be added that would have caught the original issue. (It was a pretty severe bug, surprised it wasn’t caught by any tests).

PR #63, opened 2019-05-07
Webpack optimization

mirpo:
- Updated Webpack
- Improved speed for local development
- Reduced size of dist

### DIFF
--- a/src/Connection.js
+++ b/src/Connection.js
@@ -6,16 +6,7 @@ import { ControlLayer } from 'streamr-client-protocol'
 
 const debug = debugFactory('StreamrClient::Connection')
 
-export default class Connection extends EventEmitter {
-    static get State() {
-        return {
-            DISCONNECTED: 'disconnected',
-            CONNECTING: 'connecting',
-            CONNECTED: 'connected',
-            DISCONNECTING: 'disconnecting',
-        }
-    }
-
+class Connection extends EventEmitter {
     constructor(options, socket) {
         super()
         if (!options.url) {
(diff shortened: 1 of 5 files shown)


src/Connection.js
```diff
@@ -6,7 +6,16 @@ import { ControlLayer } from 'streamr-client-protocol'
 
 const debug = debugFactory('StreamrClient::Connection')
 
-class Connection extends EventEmitter {
+export default class Connection extends EventEmitter {
+    static get State() {
```
timoxley: Does this need to be a getter? Means every *access* to `Connection.State` instantiates a new `Object`

webpack.config.js
```diff
@@ -1,17 +1,22 @@
 /* eslint-disable prefer-template */
 /* eslint-disable prefer-destructuring */
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'development' // set a default NODE_ENV
+
 const path = require('path')
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
+const webpack = require('webpack')
+const TerserPlugin = require('terser-webpack-plugin')
```
timoxley: Should this be in package.json?

src/Connection.js
mirpo: You're right, I copied idea from Subscription.js for consistency, but it was wrong idea. Updated both now.

webpack.config.js
mirpo: Yes

PR #64, opened 2019-05-09
Resend function

mirpo:
- First version of resend for testing network
- Fix send function
- Now it's possible to get raw data from publish method

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,24 +46,25 @@ const sub = client.subscribe(
 )
 ```
 
-See "Subscription options" for resend options
-
 #### Resending historical data 
 
 ```javascript
-const sub = await client.publish(
+const sub = await client.resend(
     streamId,
     {
         resend: {
             last: 5,
         },
     },
-    (message, metadata) => {
-        // This is the message handler which gets called for every incoming message in the Stream.
+    (message) => {
+        // This is the message handler which gets called for every received message in the Stream.
         // Do something with the message here!
     }
 )
 ```
+
+See "Subscription options" for resend options
+
 
 #### Programmatically creating a Stream
 
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ const sub = client.subscribe(
 
 ```javascript
 const sub = await client.resend(
-    streamId,
     {
+        stream: 'streamId',
         resend: {
             last: 5,
         },
--- a/README.md
+++ b/README.md
@@ -46,6 +46,25 @@ const sub = client.subscribe(
 )
 ```
 
+See "Subscription options" for resend options
+
+#### Resending historical data 
+
+```javascript
+const sub = await client.publish(
+    streamId,
+    {
+        resend: {
+            last: 5,
+        },
+    },
+    (message, metadata) => {
+        // This is the message handler which gets called for every incoming message in the Stream.
+        // Do something with the message here!
+    }
+)
+```
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -46,6 +46,26 @@ const sub = client.subscribe(
 )
 ```
 
+#### Resending historical data 
+
+```javascript
+const sub = await client.resend(
```
harbu: Would it make sense for consistency to pass streamId in the options object similar to `subscribe` method? I.e.,
```js
client.resend({
    stream: 'streamId',
    resend: {
        last: 5
    },
    ...
)
```

src/Connection.js
```diff
@@ -104,11 +104,15 @@ class Connection extends EventEmitter {
...
 
     send(controlLayerRequest) {
-        try {
-            this.socket.send(controlLayerRequest.serialize())
-        } catch (err) {
-            this.emit('error', err)
-        }
+        return new Promise((resolve, reject) => {
+            try {
+                this.socket.send(controlLayerRequest.serialize())
```
harbu: Should you be hooking resolve to callback? Otherwise it will resolve immediately unless `send` is sync?

E.g. from network
```js
ws.send(message, (err) => {
                            if (err) {
                                reject(err)
                            } else {
                                debug('sent to %s message "%s"', recipientAddress, message)
                                resolve()
                            }
                        })
```

Not sure if browser websocket send supports callback though.

src/StreamrClient.js
```diff
@@ -319,6 +320,21 @@ export default class StreamrClient extends EventEmitter {
...
+    async resend(streamId, options, callback) {
+        if (!streamId) {
+            throw new Error('resend: Invalid arguments: streamId is not given')
+        }
+
+        await this.ensureConnected()
+
+        const sub = new Subscription(streamId, options.partition || 0, callback, options.resend)
+
+        this._addSubscription(sub)
```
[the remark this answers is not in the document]
mirpo: Not possible now, added TODO comment

src/Connection.js
timoxley: > browser websocket send supports callback though.

it doesn't

timoxley: > unless send is sync

browser's `send` just enqueues the message to be sent, not sure there's a built-in way to detect that a specific message was received or failed to send

timoxley: This try/catch should perhaps just go specifically around the `controlLayerRequest.serialize()` part? Also not immediately sure why this is wrapped in a `Promise` since yeah, everything in here is sync

test/integration/StreamrClient.test.js
```diff
@@ -110,6 +112,136 @@ describe('StreamrClient Connection', () => {
...
+                    resend: {
+                        last: 3,
+                    },
+                },
+                (message) => {
+                    messages.push(message)
+                },
+            )
+
+            sub.on('resent', () => {
```
timoxley: for good measure, these listeners should probably all be `sub.once(`

src/Connection.js
mirpo: True, removed Promise from send function

test/integration/StreamrClient.test.js
mirpo: 👍

README.md
mirpo: Updated

PR #65, opened 2019-05-21
msgChainId: use null as indication of no value

harbu:
Use null as indication of no value instead of empty string


harbu:
In broker (fork of data-api) I need to test the presence of `publisherId` and `msgChainId` to form the Cassandra query. Currently `publisherId` is set to null when no value is present. For symmetry one would expect `msgChainId` to behave similarly.

harbu:
With respect to testing, resend behavior is currently integration tested against the data api service in `StreamrClient.test.js`. In those tests `publisherId` and `msgChainId` are both omitted. So there is test coverage in place that ensures that replacing the empty string with null is compatible with the current data api service.

PR #69, opened 2019-06-11
Swap node-cache for receptacle to get around CSP issues.

timoxley:
As described in [CORE-1696](https://streamr.atlassian.net/browse/CORE-1696), `node-cache` [uses `lodash.template`](https://github.com/mpneuried/nodecache/blob/8399b71d9a84359b3bec77fba789fc460a3fc3fa/_src/lib/node_cache.coffee#L616) which uses the function constructor, which causes [issues in under strict CSP](https://github.com/lodash/lodash/search?q=template+csp&type=Issues).

This PR swaps `node-cache` for [`receptacle`](https://github.com/DylanPiercey/receptacle), which provides similar functionality. 

* Sets limit to 10,000 cached streams.
* 30 minute TTL.
* TTL refreshed on access.

Tests pass.

Bundle size also slightly reduced: 834KiB -> 814KiB (20KiB, 2%):

### Before
```
   streamr-client.web.min.js  834 KiB       0  [emitted]  [big]  main
    streamr-client.web.min.js.map  997 KiB       0  [emitted]         main
```

### After
```
     streamr-client.web.min.js  814 KiB       0  [emitted]  [big]  main
    streamr-client.web.min.js.map  963 KiB       0  [emitted]         main
```


hpihkala:
@mthambipillai once this is landed, could you pls publish a new patch version update to npm? Could ask Peter (on the dev forum) to try the new version.

PR #71, opened 2019-06-11
Consistently use qs over querystring.

timoxley:
No point loading two packages for near identical functionality.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4248,14 +4248,14 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.0",
-          "resolved": false,
+          "resolved": "",
           "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=",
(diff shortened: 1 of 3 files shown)


timoxley:
So there is a mild difference between `qs` and `querystring` in their encoding of arrays, `qs` encodes the array as a single query component, while `querystring` encodes each array element as an individual query component with the same key:

```js
const query = {a: 3, b: [1,2,3]}
qs.stringify(query) // => "a=3&b%5B0%5D=1&b%5B1%5D=2&b%5B2%5D=3"
querystring.stringify(query) "a=3&b=1&b=2&b=3"
```
Neither is really more "correct", different serverside query parsers will support one or both of the above encodings.

This *shouldn't* affect the `listStreams` call that was changed in this PR, since none of the query parameters for listing streams are expected to be arrays, but just a FYI in case there was some undocumented or hidden magic.

PR #72, opened 2019-06-11
Add debug option. Replace console.error/warn with debug

timoxley:
Currently one cannot disable console messages in the client, whenever you get an error event, you will get an error logged to the console, regardless whether the error is handled or expected. 

This is a minor nuisance for consumers but also affects their tests: newer versions of Jest complain if console messages appear outside of test handlers. 

This PR allows opting into error logging via setting a `debug` flag in the client options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ verifySignatures | 'auto' | Determines under which conditions signed and unsigne
 autoConnect | true | If set to `true`, the client connects automatically on the first call to `subscribe()`. Otherwise an explicit call to `connect()` is required.
 autoDisconnect | true  | If set to `true`, the client automatically disconnects when the last stream is unsubscribed. Otherwise the connection is left open and can be disconnected explicitly by calling `disconnect()`.
 maxPublishQueueSize | 10000 | Only in effect when `autoConnect = true`. Controls the maximum number of messages to retain in internal queue when client has disconnected and is reconnecting to Streamr.
+debug | false | Enable debug logging
 
 ### Authentication options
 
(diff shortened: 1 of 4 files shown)


hpihkala:
- At the end of the `README`, there's a section about debug logging, which seems affected by the changes but it has not been updated.
- If I'm understanding correctly, now the user would have to do 2 things to enable debug logging: give `debug: true` in the options AND set the env variable (node) / `localStorage.debug` (browser) setting. Isn't that a bit redundant? Will lead to confusion when they set `debug: true` and nothing appears?

So far the paradigm has been to use `debug` for debug-level (detailed) logging, which can be toggled on/off locally via the logger name system, and print more serious things like errors (which we need people to see) to the console without ability to turn them off. Is there something wrong with the approach? Leveling everything, including errors, to be debug-level logging doesn't sound too good - is there a better way?

timoxley:
> If I'm understanding correctly, now the user would have to do 2 things to enable debug logging

Both work, kind of two ways of doing the same thing, but the `debug` option makes the logging explicit.

> So far the paradigm has been to use debug for debug-level (detailed) logging …
>  is there a better way?

@hpihkala Library code generally shouldn't be logging anything to the console, and if it does there needs to be a way to turn it off.

One option that preserves the separation between debug-level logging and error logging could be to only trigger the `console.error` lines if the client's `debug` option is `true`. This may even be a better option than swapping the console logs to `debug`.

> At the end of the README, there's a section about debug logging, which seems affected by the changes but it has not been updated.

Good point, missed that. Updated the `debug` option to point at the `debug` module and mentioned the env var in 230c1ae7, but I'm thinking perhaps the above option is actually better.

timoxley:
Closing for now, going to replace `console`/`debug` with something better that supports multiple log levels.

PR #73, opened 2019-06-14
expire list of publishers after 30 minutes

mthambipillai:
The list of publishers is used to determine the validity of signatures. It should be fetched again periodically so that signatures from new publishers are accepted and signatures from old publishers are rejected.


mthambipillai:
Unit test added. I think it's better to leave it as an internal not configurable parameter since signature verification itself is kind of internal.

PR #74, opened 2019-07-02
Net 10 ordering util

mthambipillai:
Refactoring of the resend and gap filling logic to use an `OrderingUtil`:
`unorderedMessages -------> | OrderingUtil | -------> orderedMessages`

The goal for this refactoring is to make this utility available to the code of the network node where it will be used as well and avoid code duplication.

For now, this utility is part of the `streamr-client` repo. We haven't really decided where this code should go, it can be decided and moved if necessary in a subsequent PR.
Several options:
- keep it in `streamr-client` and add a dependency to `streamr-client` in `network`
- add it to the utilities in `streamr-client-protocol` which is anyway a dependency of `network`
- create a new `streamr-utils` repo, add it there and add a dependency to `streamr-utils` in `network`


src/OrderingUtil.js
```diff
@@ -0,0 +1,121 @@
+import debugFactory from 'debug'
+import { Errors } from 'streamr-client-protocol'
+
+const debug = debugFactory('StreamrClient::OrderingUtil')
+
+export default class OrderingUtil {
+    constructor(streamId, streamPartition, inOrderHandler, gapHandler, options) {
```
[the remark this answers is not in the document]
mthambipillai: I don't have a preference. @hpihkala what do you think?

src/Subscription.js
```diff
@@ -52,6 +45,24 @@ class Subscription extends EventEmitter {
         if (this.resendOptions.from == null && this.resendOptions.to != null) {
             throw new Error('"from" must be defined as well if "to" is defined.')
         }
+        this.initialResendDone = Object.keys(this.resendOptions).length === 0
+        this.state = Subscription.State.unsubscribed
+        this.groupKeys = groupKeys || {}
+        this.queue = []
```
hpihkala: - Messages in the initial resend should be processed as soon as possible (correct order is reached), instead of buffering the whole resend into memory and starting to process messages when the resend is complete
- Would be nice to avoid having the queuing and queue processing code in this class

Could the `OrderingUtil` be used for the initial resend as well?

What if `orderingUtil.add(msg, forceQueue)` takes a second parameter which is used to force the realtime messages into the queue until the initial resend is complete?

src/OrderingUtil.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    async add(unorderedStreamMessage) {
+        const key = unorderedStreamMessage.getPublisherId() + unorderedStreamMessage.getMsgChainId()
+        if (!this._checkForGap(unorderedStreamMessage.prevMsgRef, key)) { // if it is the next message, process it
+            this._processMsg(key, unorderedStreamMessage)
+            await this._checkQueue()
+        } else {
+            if (!this.gaps[key]) { // send a gap fill request only if there is not one already in progress
+                this._scheduleGapFillRequests(key, unorderedStreamMessage)
```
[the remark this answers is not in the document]
hpihkala: I guess it's intentional, as you're sort of scheduling a check for each message, but another approach would be to put the messages in the queue and then schedule (one) check for the queue.

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+
+    clearGap() {
+        clearInterval(this.gap)
+        this.gap = undefined
+    }
+
+    _isNextMessage(unorderedStreamMessage) {
+        const isFirstMessage = this.lastReceivedMsgRef === null
+        return isFirstMessage
+            || (this.lastReceivedMsgRef !== null
```
hpihkala: Condition `this.lastReceivedMsgRef !== null` is unnecessary, because we already know that `isFirstMessage` is false?

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+                && unorderedStreamMessage.prevMsgRef !== null
```
hpihkala: What if the messages are NOT chained (i.e. `unorderedStreamMessage.prevMsgRef` is `null`). These messages should always be considered to be a valid next message (unless they are older than the newest processed message and fail the `compareTo` check, which is done earlier).

Is there test coverage for unchained messages?

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+                && unorderedStreamMessage.prevMsgRef.compareTo(this.lastReceivedMsgRef) === 0)
+    }
+
+    _insertInOrderedQueue(unorderedStreamMessage) {
+        this.queue.push(unorderedStreamMessage)
+    }
+
+    _getTopMsgInQueue() {
```
hpihkala: Maybe unnecessary one-liner function?

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+        return this.queue.peek()
+    }
+
+    _popQueue() {
```
hpihkala: Maybe unnecessary one-liner function?

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+        return this.queue.pop()
+    }
+
+    _checkQueue() {
+        const msg = this._getTopMsgInQueue()
+        if (msg && this._isNextMessage(msg)) {
+            this._popQueue()
+            this.clearGap()
```
hpihkala: Could use a comment here to explain the logic, something along the lines of:
> If the next message is found in the queue, any gap must have been filled, so clear the timer

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+            this._process(msg)
```
hpihkala: This recursion will lead to a stack overflow if the queue is long, and is less efficient than iteration due to the many function calls. Please replace recursion with iteration.

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+        const msgRef = unorderedStreamMessage.getMessageRef()
+        if (this.lastReceivedMsgRef && msgRef.compareTo(this.lastReceivedMsgRef) < 0) {
+            // Prevent double-processing of messages for any reason
+            debug('Already received message: %o, lastReceivedMsgRef: %d. Ignoring message.', msgRef, this.lastReceivedMsgRef)
+        }
+
+        if (this._isNextMessage(unorderedStreamMessage)) {
+            this._process(unorderedStreamMessage)
+        } else {
+            if (!this.gap) {
```
hpihkala: Is it possible for `setInterval` to return a falsey value? Changing the comparison to `this.gap != null` would unburden the reader from that thought

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+    _process(msg) {
+        this.lastReceivedMsgRef = msg.getMessageRef()
+        this.inOrderHandler(msg)
+        this._checkQueue()
+    }
+
+    _scheduleGap() {
+        this.gapRequestCount = 0
+        this.gap = setInterval(() => {
+            const from = this.lastReceivedMsgRef == null ? this._getTopMsgInQueue().prevMsgRef
```
hpihkala: We should not end up here unless `this.lastReceivedMsgRef != null`, because a gap cannot be detected from the first message!

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,108 @@
...
+                : new MessageRef(this.lastReceivedMsgRef.timestamp, this.lastReceivedMsgRef.sequenceNumber + 1)
+            const to = this._getTopMsgInQueue().prevMsgRef
+            if (this.gapRequestCount < MAX_GAP_REQUESTS) {
+                this.gapRequestCount += 1
+                this.gapHandler(from, to, this.publisherId, this.msgChainId)
+            } else {
+                this.clearGap()
+                throw new Error(
```
hpihkala: Could throw a specialized subclass of Error to enable the client/app to handle better this particular error. 

- The error should contain the `from` and `to` as fields on the object, not only in the error message: `throw new GapFillFailedError(msg, from, to)`
- Note that `publisherId` and `msgChainId` are included in the `from` and `to` msgRefs, right? So they are unnecessary in the error (and error message).

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,101 @@
...
+        this.gap = undefined
+    }
+
+    _isNextMessage(unorderedStreamMessage) {
+        const isFirstMessage = this.lastReceivedMsgRef === null
+        return isFirstMessage
+            // is chained and next
+            || (unorderedStreamMessage.prevMsgRef !== null && unorderedStreamMessage.prevMsgRef.compareTo(this.lastReceivedMsgRef) === 0)
+            // is unchained and newer
+            || (unorderedStreamMessage.prevMsgRef === null && unorderedStreamMessage.getMessageRef().compareTo(this.lastReceivedMsgRef) >= 0)
```
harbu: Should this be strict `unorderedStreamMessage.getMessageRef().compareTo(this.lastReceivedMsgRef) > 0`?

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,101 @@
...
+        this.gapRequestCount = 0
+        this.gap = setInterval(() => {
+            const from = new MessageRef(this.lastReceivedMsgRef.timestamp, this.lastReceivedMsgRef.sequenceNumber + 1)
+            const to = this.queue.peek().prevMsgRef
+            if (this.gapRequestCount < MAX_GAP_REQUESTS) {
+                this.gapRequestCount += 1
+                this.gapHandler(from, to, this.publisherId, this.msgChainId)
+            } else {
+                this.clearGap()
+                throw new GapFillFailedError(from, to, this.publisherId, this.msgChainId, MAX_GAP_REQUESTS)
```
harbu: Hmm.. is this exception effectively catchable by the user of this class since it is being thrown inside `setInterval`? Maybe invoking an error handler callback would work?

src/OrderedMsgChain.js
```diff
@@ -0,0 +1,101 @@
...
+            this._checkQueue()
+        } else {
+            if (!this.gap) {
+                this._scheduleGap()
+            }
+            this.queue.push(unorderedStreamMessage)
+        }
+    }
+
+    addError(err) {
```
harbu: I think this class doesn't necessarily need to know about the structure of the error. You could dig out the streamMessage from the error outside this class and then refactor this to be something like
```js
markMessageExplicitly(streamMessage) {
    if (this._isNextMessage(streamMessage)) {
        this.lastReceivedMsgRef = err.getMessageRef()
    }
}
```

src/OrderingUtil.js
```diff
@@ -0,0 +1,38 @@
...
+
+    _getChain(publisherId, msgChainId) {
+        const key = publisherId + msgChainId
+        if (!this.orderedChains[key]) {
+            this.orderedChains[key] = new OrderedMsgChain(publisherId, msgChainId, this.inOrderHandler, this.gapHandler, this.gapFillTimeout)
+        }
+        return this.orderedChains[key]
+    }
+
+    addError(err) {
```
harbu: Same comment as before in `OrdererMsgChain`. No need for this class to be tied to the structure of the error.

src/Subscription.js
```diff
@@ -327,14 +238,7 @@ class Subscription extends EventEmitter {
...
          */
-        let key
-        if (err.streamMessage) {
-            key = err.streamMessage.getPublisherId() + err.streamMessage.messageId.msgChainId
-        }
-
-        if (err instanceof Errors.InvalidJsonError && !this.checkForGap(err.streamMessage.prevMsgRef, key)) {
-            this.lastReceivedMsgRef[key] = err.streamMessage.getMessageRef()
-        }
+        this.orderingUtil.addError(err)
```
harbu: Related to the `addError` comments above. Move the digging of streamMessage from error maybe here?

test/unit/OrderedMsgChain.test.js
```diff
@@ -0,0 +1,157 @@
...
+            done()
+        }
+    })
+    it('handles unordered messages in order (large randomized test)', () => {
+        const expected = [msg1]
+        let i
+        for (i = 2; i <= 1000; i++) {
+            expected.push(createMsg(i, 0, i - 1, 0))
+        }
+        const shuffled = shuffle(expected)
```
harbu: It would be a good idea to provide the `shuffle` function a fixed seed if possible so that the test it deterministic.

src/OrderedMsgChain.js
mthambipillai: Yes it is catchable. There is a test that wraps the entire class in a `try` `catch` block and effectively catches the error.

test/unit/OrderedMsgChain.test.js
mthambipillai: Why does the test need to be deterministic?
I made it random on purpose: I thought that having a different input each time the test is run makes it way more robust than a deterministic test. No matter the order of the input messages, we always expect the same order of output messages

src/OrderedMsgChain.js
harbu: It think it is catchable in the test because it uses fake timers which seem to invoke timeout callbacks synchronously. Try running
```js
const sinon = require('sinon')

function throwErrorInTimeout() {
  setTimeout(() => {
    throw new Error('this is an error')
  }, 10)
}

it('without fake timers', (done) => {
  try {
    throwErrorInTimeout()
  } catch (e) {
    console.error('caught' + e)
    done()
  }
})

it('with fake timers', (done) => {
  const clock = sinon.useFakeTimers()
  try {
    throwErrorInTimeout()
    clock.tick(10)
  } catch (e) {
    console.error('caught' + e)
    clock.restore()
    done()
  }
})
```

test/unit/OrderedMsgChain.test.js
harbu: Reproducibility of failing tests is important. Let's say I'm working on something completely different, run all tests, and then this test, on some specific outcome of shuffle, happens to fail. It won't really help me fix the issue because the next run of this test passes / behaves differently.

Pretty good discussion about this here: https://softwareengineering.stackexchange.com/questions/221632/testing-deterministic-or-non-deterministic

Basically what it boils down to is that tests needs to be reproducible. And that requires the inputs to the process to be known. Some ideas on how this can be achieved
1) use a fixed seed
2) try running multiple times with a set of random pre-selected seeds
3) console.log the order used and make it easy for the user to run the same scenario using that order

PR #75, opened 2019-07-04
Valid publishers optimization

mthambipillai:
The logic to check if an Ethereum address is a valid publisher is now the following:
- check cache (maps publisherId --> boolean)
- if hit, return cache result
- if miss, call the new endpoint `streams/id/publisher/address`, cache the result and return it

### DIFF
--- a/src/SubscribedStream.js
+++ b/src/SubscribedStream.js
@@ -10,7 +10,7 @@ export default class SubscribedStream {
     }
 
     async getPublishers() {
-        if (!this.publishersPromise || (Date.now() - this.lastAccess) > PUBLISHERS_EXPIRATION_TIME) {
+        if (!this.publishersPromise || (Date.now() - this.lastUpdated) > PUBLISHERS_EXPIRATION_TIME) {
             this.publishersPromise = this._client.getStreamPublishers(this.streamId).then((publishers) => {
                 const map = {}
                 publishers.forEach((p) => {
@@ -18,12 +18,12 @@ export default class SubscribedStream {
                 })
                 return map
             })
-            this.lastAccess = Date.now()
+            this.lastUpdated = Date.now()
         }
         return this.publishersPromise
     }
 
-    async isPublisher(publisherId) {
+    async _isPublisher(publisherId) {
         if (!this.isPublisherPromises[publisherId]) {
             this.isPublisherPromises[publisherId] = this._client.isStreamPublisher(this.streamId, publisherId)
         }
@@ -35,7 +35,7 @@ export default class SubscribedStream {
         if (cache[publisherId]) {
             return cache[publisherId]
         }
-        const isValid = await this.isPublisher(publisherId)
+        const isValid = await this._isPublisher(publisherId)
         cache[publisherId] = isValid
         return isValid
     }
--- a/src/SubscribedStream.js
+++ b/src/SubscribedStream.js
@@ -6,21 +6,48 @@ export default class SubscribedStream {
         this._client = client
         this.streamId = streamId
         this.subscriptions = {}
+        this.isPublisherPromises = {}
     }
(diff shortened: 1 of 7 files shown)


src/SubscribedStream.js
```diff
@@ -6,30 +6,60 @@ export default class SubscribedStream {
         this._client = client
         this.streamId = streamId
         this.subscriptions = {}
+        this.isPublisherPromises = {}
     }
 
-    getPublishers() {
+    async getPublishers() {
         if (!this.publishersPromise || (Date.now() - this.lastAccess) > PUBLISHERS_EXPIRATION_TIME) {
```
harbu: If a cache gets hit on a regular basis (that is `<= PUBLISHERS_EXPIRATION_TIME`), it means that the `this._client.getStreamPublishers`will not be invoked? Would it be better to have time-based expiration so that we have a guarantee that it gets updated every now and then regardless of access patterns?

src/SubscribedStream.js
```diff
@@ -6,30 +6,60 @@ export default class SubscribedStream {
...
     }
 
+    async isPublisher(publisherId) {
+        if (!this.isPublisherPromises[publisherId]) {
+            this.isPublisherPromises[publisherId] = this._client.isStreamPublisher(this.streamId, publisherId)
+        }
+        return this.isPublisherPromises[publisherId]
+    }
+
+    async isValidPublisher(publisherId) {
```
harbu: What is the difference between `isPublisher` and `isValidPublisher` ?

src/SubscribedStream.js
```diff
@@ -6,30 +6,60 @@ export default class SubscribedStream {
...
+        const cache = await this.getPublishers()
+        if (cache[publisherId]) {
```
harbu: Isn't the code from here onwards in this function basically the same logic as what `isPublisher` does? As `cache == this.isPublisherPromises`?

src/SubscribedStream.js
mthambipillai: The cache gets updated every `PUBLISHERS_EXPIRATION_TIME` milliseconds in any case because `this.lastAccess` is updated only inside this `if` statement, not for every cache hit. But maybe this variable should be renamed to `this.lastUpdated` to avoid confusion.

src/SubscribedStream.js
mthambipillai: `isPublisher` calls the endpoint but stores the promise to avoid multiple calls to the endpoint when the function is called again before the promise is resolved.
`isValidPublisher` makes use of the 2 endpoints.

src/SubscribedStream.js
mthambipillai: `cache == this.publishersPromise` but `cache != this.isPublisherPromises`.
The logic is different: it uses the first endpoint to get the full list and periodically refresh it and uses the second endpoint when there is a cache miss.

src/SubscribedStream.js
harbu: Ah indeed true. Rename would probably be good idea.

src/SubscribedStream.js
harbu: Ah my bad. I mixed them up and thought there is only one variable.

src/SubscribedStream.js
harbu: If I understand correctly, the intention is that a user of this class would use `isValidPublisher` instead of `isPublisher`? If this is true, would it make sense to rename `isPublisher` to `_isPublisher` to communicate that it is internal in nature. Or is there a case where a user of this class would want to use `isPublisher` instead of `isValidPublisher`?

mthambipillai: It is internal so yes I can prepend its name with `_`

PR #82, opened 2019-07-08
Fix tests and code to work without --forceExit

mirpo:


test/integration/StreamrClient.test.js
```diff
@@ -665,7 +666,7 @@ describe('StreamrClient', () => {
                         done()
                     })
                 })
-            }, 10000)
+            }, TIMEOUT * 0.8)
```
harbu: Why the multiplier 0.8? Just change value of `TIMEOUT`?

PR #83, opened 2019-07-19
Key exchange

mthambipillai:
This PR adds the key exchange mechanism. Contains:
- endpoints to get stream subscribers and check if address is a subscriber
- storage of group key histories (to be able to answer key requests)
- sending of group key requests when decryption failures occur
- handling of these requests and sending of group key responses
- handling of these responses to update group keys and decrypt messages queued in between

This PR is quite big so in addition to the commit history, the following reviewing steps could be useful:
1) `EncryptionUtil`: additional methods for RSA
2) `StreamEndpoints`: 2 new methods for the subscribers endpoints
3) `KeyStorageUtil` `KeyHistoryStorageUtil`, `LatestKeyStorageUtil` and `GroupKeyHistory`: utility to store the keys used by the publisher to encrypt messages and several methods to get/add keys.
4) `MessageCreationUtil`: additional methods to create group key requests and responses.
5) `KeyExchangeUtil`: contains handlers called when receiving group key requests and responses. Makes use of the classes above.
6) `Subscription`: changes to the logic to request group keys when decryption fails.
7) `StreamrClient`: mainly inbox stream subscription to call the handlers in `KeyExchangeUtil`

Notes:
- This PR does not include any revocation / rekeying mechanism yet.
- Node version had to be upgraded to 10 to be able to use the RSA encryption/decryption functions from the standard `crypto` library.


hpihkala:
@mthambipillai I have no chance of reviewing this before my holiday, plus would be good to get a third pair of eyes on this. Please introduce this to @harbu and you guys do a first round of review over the next two weeks. I'll jump in when I get back.

mthambipillai:
@harbu we could have a call one of these days so that I can fill you in with the big picture, it will be easier to review then.

timoxley:
@mthambipillai can you rebase on master and push? I'll review this and we can try get this landed.

mthambipillai:
@timoxley yes I'll resolve the conflicts. Anyway, first the corresponding Java client PR must be reviewed as this might result in replicating changes in this PR as well.

hpihkala:
Alright so let's get this big PR reviewed! It will need more than one pair of eyes, so I'd suggest the following:

- @harbu has already reviewed the Java client PR which adds the same new features, so you already have the background and insight into what's going on here. Leveraging that familiarity, please do a pass on this PR as well, focusing perhaps more on the correctness of functionality (as opposed to JS best practices etc.).
- @timoxley could I ask you to also review the changes, complementing Eric's review angle by perhaps focusing more on the JS side of things (as opposed to correctness of functionality). You're welcome to cover the functionality perspective too, although it might require some background info first. @mthambipillai can provide a walkthrough if required.

src/StreamrClient.js
```diff
@@ -110,10 +134,49 @@ export default class StreamrClient extends EventEmitter {
...
+                                )
+                            }
+                        }
+                    } else if (streamMessage.contentType === StreamMessage.CONTENT_TYPES.ERROR_MSG) {
+                        debug('WARN: Received error of type %s from %s: %s',
+                            streamMessage.getParsedContent().code, streamMessage.getPublisherId(), streamMessage.getParsedContent().message)
+                    } else {
+                        throw new InvalidContentTypeError(`Cannot handle message with content type: ${streamMessage.contentType}`)
+                    }
+                } catch (err) {
```
timoxley: This catches a lot, might want to verify the type of the error and rethrow non-matching.

src/StreamrClient.js
```diff
@@ -441,18 +509,37 @@ export default class StreamrClient extends EventEmitter {
...
         if (options.groupKeys) {
-            this.options.subscriberGroupKeys[options.stream] = options.groupKeys
+            Object.keys(options.groupKeys).forEach((publisherId) => {
+                EncryptionUtil.validateGroupKey(options.groupKeys[publisherId])
+                if (!this.options.subscriberGroupKeys[options.stream]) {
+                    this.options.subscriberGroupKeys[options.stream] = {}
+                }
+                this.options.subscriberGroupKeys[options.stream][publisherId] = {
+                    groupKey: options.groupKeys[publisherId],
+                    start: Date.now()
```
timoxley: it's possible these will get different values for `start`. Might not matter, but perhaps should do `const now = Date.now()` *outside* the `forEach` loop so they all get the same `start`?

src/RealTimeSubscription.js
```diff
@@ -25,6 +29,25 @@ export default class RealTimeSubscription extends AbstractSubscription {
...
+    // passing publisherId separately to ensure it is lowercase (See call of this function in AbstractSubscription.js)
+    _decryptOrRequestGroupKey(msg, publisherId) {
+        try {
+            const newGroupKey = EncryptionUtil.decryptStreamMessage(msg, this.groupKeys[publisherId])
+            delete this.alreadyFailedToDecrypt[publisherId]
+            if (newGroupKey) {
+                this.groupKeys[publisherId] = newGroupKey
+            }
+            return true
+        } catch (e) {
```
timoxley: Best practice to only put try/catch only around code which you're expecting to fail. 

```js
let newGroupKey
try {
  newGroupKey = EncryptionUtil.decryptStreamMessage(msg, this.groupKeys[publisherId])
} catch (e) {
  // …
}
// happy path
delete this.alreadyFailedToDecrypt[publisherId]
// etc
```

src/StreamrClient.js
```diff
@@ -110,10 +134,49 @@ export default class StreamrClient extends EventEmitter {
...
+                            }
+                        }
+                    } else if (streamMessage.contentType === StreamMessage.CONTENT_TYPES.ERROR_MSG) {
+                        debug('WARN: Received error of type %s from %s: %s',
+                            streamMessage.getParsedContent().code, streamMessage.getPublisherId(), streamMessage.getParsedContent().message)
+                    } else {
+                        throw new InvalidContentTypeError(`Cannot handle message with content type: ${streamMessage.contentType}`)
+                    }
+                } catch (err) {
+                    debug('WARN: %s', err.message)
```
timoxley: e.g. if user doesn't have `debug` enabled, and no `streamMessage.signature`, any errors will be totally suppressed, which could be very frustrating to track down.

src/StreamrClient.js
```diff
@@ -702,12 +797,31 @@ export default class StreamrClient extends EventEmitter {
...
+        if (!this.options.subscriberGroupKeys[streamId]) {
+            this.options.subscriberGroupKeys[streamId] = {}
+        }
+        const last = groupKeys[groupKeys.length - 1]
+        const current = this.options.subscriberGroupKeys[streamId][publisherId]
+        if (!current || last.start > current.start) {
+            this.options.subscriberGroupKeys[streamId][publisherId] = last
+        }
+        // TODO: fix this hack in other PR
+        this.subscribedStreamPartitions[streamId + '0'].setSubscriptionsGroupKeys(publisherId, groupKeys.map((obj) => obj.groupKey))
```
timoxley: does this hack still need a fix?

test/unit/EncryptionUtil.test.js
```diff
@@ -0,0 +1,171 @@
...
+import EncryptionUtil from '../../src/EncryptionUtil'
+
+const { StreamMessage } = MessageLayer
+
+describe('EncryptionUtil', () => {
+    it('rsa decryption after encryption equals the initial plaintext', () => {
+        const encryptionUtil = new EncryptionUtil()
+        const plaintext = 'some random text'
+        const ciphertext = EncryptionUtil.encryptWithPublicKey(Buffer.from(plaintext, 'utf8'), encryptionUtil.getPublicKey())
+        assert.deepStrictEqual(encryptionUtil.decryptWithPrivateKey(ciphertext).toString('utf8'), plaintext)
```
timoxley: Should use `jest`'s `expect` rather than `assert`: https://jestjs.io/docs/en/expect

test/unit/EncryptionUtil.test.js
```diff
@@ -0,0 +1,171 @@
...
+                format: 'pem',
+            },
+        })
+        assert.throws(() => {
+            // eslint-disable-next-line no-new
+            new EncryptionUtil({
+                privateKey: keys.privateKey,
+                publicKey: 'wrong public key',
+            })
+        }, /Error/)
```
timoxley: This `/Error/` validation.doesn't do much since almost every Error will match `/Error/`. Perhaps should update it to match against the actual expected error, or do nothing instead.

src/StreamrClient.js
mthambipillai: Yes this is not related to key-exchange and should be done once this PR lands.

src/StreamrClient.js
mthambipillai: I agree, but this is consistent to other message error handling in `StreamrClient.js`. So I think it's better to rethink after this PR at a global level how errors on reception of external messages should be handled/logged.

timoxley:
Tests currently failing due to unrelated issue, which should be fixed by https://github.com/streamr-dev/streamr-client-javascript/pull/126

src/KeyExchangeUtil.js
```diff
@@ -0,0 +1,131 @@
...
+    async getSubscribers(streamId) {
+        if (!this.subscribersPromise || (Date.now() - this.lastAccess) > SUBSCRIBERS_EXPIRATION_TIME) {
+            this.subscribersPromise = this._client.getStreamSubscribers(streamId).then((subscribers) => {
+                const map = {}
+                subscribers.forEach((s) => {
+                    map[s] = true
+                })
+                return map
+            })
+            this.lastAccess = Date.now()
```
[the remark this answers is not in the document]
mthambipillai: I think this is solved by `this.subscribersPromise`: on a second call before it is resolved, it will just return the same promise.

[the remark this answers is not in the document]
harbu: Ahh.. my bad. I read the curly brackets wrong. I read that `this.lastAccess = Date.now()` is inside `then(() => {..})` even though it is outside.

PR #84, opened 2019-07-23
CORE-1764: handle case of date timestamp

mthambipillai:


src/StreamrClient.js
```diff
@@ -287,9 +287,10 @@ export default class StreamrClient extends EventEmitter {
             throw new Error(`First argument must be a Stream object or the stream id! Was: ${streamObjectOrId}`)
         }
 
+        const timestampAsNumber = typeof timestamp === 'number' ? timestamp : timestamp.getTime()
```
hpihkala: It would probably be good practice to "prefer" `Date` objects, although we should maintain backwards compatibility. What do you think if we spec the `timestamp` argument as follows:

- A `Date` object, or a valid argument to the `Date` constructor `new Date(...)`

This would actually also support ISO8601 strings out of the box. Then in our code:

```
async publish(streamObjectOrId, data, timestamp = new Date(), partitionKey = null, groupKey) {
....
    const timestampAsNumber = timestamp instanceof Date ? timestamp.getTime() : new Date(timestamp).getTime()
```

- The `README.md` should be updated to cover all the arguments to `publish`.

PR #86, opened 2019-08-02
Refactor subscription

mthambipillai:
The `Subscription` class was becoming quite complex because for every message it needed to handle it differently according to flags like `isInitialResendDone` and `isResend`. For the key exchange mechanism, this led to even more complexity: https://github.com/streamr-dev/streamr-client-javascript/blob/key-exchange/src/Subscription.js. This PR refactors the class into 4 new and more manageable classes:
- `AbstractSubscription`: common functionality for every kind of subscription
- `HistoricalSubscription`: has a clear start and end. Used directly in the explicit `resend()` function.
- `RealTimeSubscription`: starts now and the only resends are for gap filling. Used directly when calling `subscribe()` with no resend options.
- `CombinedSubscription`: Used when calling `subscribe()` with resend options. It wraps another subscription, first a `HistoricalSubscription` for the resend options and queues real-time messages. When it's done, it wraps a `RealTimeSubscription`, starts it with the last received message reference during the initial resend and then feeds it the queued messages.

This refactoring is only internal and doesn't change anything to the logic, but it gets rid of the control flags in the former `Subscription` class and will make the code for the key exchange mechanism much cleaner. This is because the way the keys are requested is different whether the messages are historical or whether they are real-time.


src/CombinedSubscription.js
```diff
@@ -0,0 +1,74 @@
...
+        this.sub = new HistoricalSubscription(streamId, streamPartition, callback, options, groupKeys, this.propagationTimeout, this.resendTimeout)
+        this.sub.on('resend done', async (lastReceivedMsgRef) => {
+            const realTime = new RealTimeSubscription(
+                streamId, streamPartition, callback, groupKeys, this.propagationTimeout, this.resendTimeout, lastReceivedMsgRef
+            )
+            await Promise.all(this.sub.realTimeMsgsQueue.map((msg) => realTime.handleBroadcastMessage(msg, () => true)))
+            this.sub.stop()
+            this.sub = realTime
+        })
+        this.sub.on('done', () => this.emit('done'))
```
harbu: Should any of these event listeners be bound to  `realTime` when the switch from historical to realtime happens?

src/CombinedSubscription.js
```diff
@@ -0,0 +1,74 @@
+import HistoricalSubscription from './HistoricalSubscription'
+import RealTimeSubscription from './RealTimeSubscription'
+import AbstractSubscription from './AbstractSubscription'
+
+export default class CombinedSubscription extends AbstractSubscription {
```
harbu: Do we need to `extend`? This class seems to be mostly about delegating to `this.sub` and not using many of the facilities provided by the base class?

src/HistoricalSubscription.js
```diff
@@ -0,0 +1,113 @@
...
+            this.setState(AbstractSubscription.State.unsubscribed)
+            this._clearGaps()
+        })
+
+        this.on('error', () => {
+            this._clearGaps()
+        })
+    }
+
+    async handleBroadcastMessage(msg, verifyFn) {
```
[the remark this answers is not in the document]
harbu: Ah I think understand know, since this may be part of `CombinedSubscription` we need to store the messages so when we switch to RealTimeSubscription we can re-play them.

src/HistoricalSubscription.js
```diff
@@ -0,0 +1,113 @@
...
+        await AbstractSubscription.validate(msg, verifyFn)
+        this.emit('message received')
+        this.realTimeMsgsQueue.push(msg)
```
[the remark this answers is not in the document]
harbu: Idea: what if instead of storing the messages in a queue here (which may or may not be used later), we instead emit an event. Then `CombinedSubscription` could take care of the queue and add messages to it by listening to this event. And when not using `CombinedSubscription`, we can simply choose to ignore the event?

src/AbstractSubscription.js
```diff
@@ -0,0 +1,125 @@
...
+            return await fn()
+        } catch (err) {
+            console.error(err)
+            this.emit('error', err)
+            // Swallow rejection
+            return Promise.resolve()
+        }
+    }
+
+    static async validate(msg, verifyFn) {
```
harbu: This seems a bit off here, perhaps. Could this logic be moved into the actual `verifyFn` which seems to be primarily `stream.verifyStreamMessage`?

mthambipillai:
The main reason why `CombinedSubscription` inherits `AbstractSubscription` is because the `subscribe` method returns a subscription. So the keep the same public facing API, the returned `RealTimeSubscription` and `CombinedSubscription` should have a common parent class.
In Javascript this is not strictly necessary but it will be in the Java client.

src/HistoricalSubscription.js
```diff
@@ -0,0 +1,113 @@
...
+            throw new Error('"from" must be defined as well if "to" is defined.')
+        }
+
+        /** * Message handlers ** */
+
+        this.on('unsubscribed', () => {
+            this._clearGaps()
+        })
+
+        this.on('disconnected', () => {
```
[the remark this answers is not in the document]
mthambipillai: Couldn't either.. That code is prior to this PR. I only see in tests that it is manually triggered with `sub.emit('disconnected')`. Could the goal be that the user of the library has the possibility to emit this event from the subscription object returned by `subscribe()`?

src/RealTimeSubscription.js
```diff
@@ -0,0 +1,123 @@
...
+        this.orderingUtil.lastReceivedMsgRef = startMsgRef
+
+        /** * Message handlers ** */
+
+        this.on('unsubscribed', () => {
+            this._clearGaps()
+            this.setResending(false)
+        })
+
+        this.on('disconnected', () => {
```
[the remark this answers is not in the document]
mthambipillai: same answer as above, don't know if this is in case the user manually emits the event on the subscription event

src/HistoricalSubscription.js
```diff
@@ -0,0 +1,113 @@
...
+            this.setState(AbstractSubscription.State.unsubscribed)
+            this._clearGaps()
+        })
+
+        this.on('error', () => {
```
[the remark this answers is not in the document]
mthambipillai: This was also already implemented before this PR. I believe it is intentional yes.

src/AbstractSubscription.js
mthambipillai: Why doesn't it belong here? This code was in `Subscription.js`, I just moved it in its separate function to avoid duplication since it is used in two different methods.
Moving this code to `verifyFn` would increase the size of `StreamrClient.js` which is already quite big.

harbu:
In Java I would recommend having five files:

- Interface `interface Subscription { ...  }` that acts as the general contract for subscriptions
- Abstract base class `abstract class AbstractSubscription implements Subscription { ... }` that contains the shared common behavior of realtime and historical subscriptions
- `class HistoricalSubscription extends AbstractSubscription`
- `class RealTimeSubscription extends AbstractSubscription`
- `class CombinedSubscription implements Subscription`

This way it is clear to the reader immediately that `CombinedSubscription` isn't using the shared facilities of `AbstractSubscription` to implement its behavior.

src/AbstractSubscription.js
harbu: The last statements of the function seem fine to me. What I have trouble understanding are the first two checks.

```js
if (msg.version !== 31) {
     throw new Error(`Can handle only StreamMessageV31, not version ${msg.version}`)
}
```
There doesn't seem to be anything inherit about this function that makes it incompatible with other versions. Isn't it the `verifyFn` function itself the one who knows which versions it is or is not compatible with? What if we were passing this function an alternative `verifyFn` that _is_ compatible with other versions? Why would we want to prevent its usage here?

```js
if (msg.prevMsgRef == null) {
    debug('handleMessage: prevOffset is null, gap detection is impossible! message: %o', msg)
}
```
Isn't gap detection done mostly in `OrderingUtil`? If I saw this log message, I would expect to find the emitting debug statement somewhere in `OrderingUtil` or its related classes.

test/unit/HistoricalSubscription.test.js
```diff
@@ -0,0 +1,550 @@
...
+                    timestamp,
+                    timestamp === 1 ? 0 : timestamp - 1,
+                ))
+
+                const received = []
+
+                const sub = new HistoricalSubscription(msg.getStreamId(), msg.getStreamPartition(), (content, receivedMsg) => {
+                    received.push(receivedMsg)
+                    if (received.length === 5) {
+                        assert.deepEqual(msgs, received)
```
harbu: I would add a `done()` in here to be extra sure that the callbacks (and hence this assertion) were indeed invoked.

PR #91, opened 2019-09-04
Add missing babel-runtime dependency.

timoxley:
Moved `babel-runtime` from `devDependencies` to `dependencies`.

When using the client from an npm installation it fails because of a missing dependency on `babel-runtime`. `streamr-client@2.2.1` works. Looks like this exact issue occurred before and was fixed in 1419cc63 but regressed in 5c2ff80c. It's easy to miss this as we don't run tests against the built code with no dev dependencies installed.

e.g. Go here and hit "Run":  https://npm.runkit.com/streamr-client 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1064,13 +1064,15 @@
           "version": "0.3.2",
           "resolved": "https://registry.npmjs.org/array-unique/-/array-unique-0.3.2.tgz",
           "integrity": "sha1-qJS3XUvE9s1nnvMkSp/Y9Gri1Cg=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "braces": {
           "version": "2.3.2",
           "resolved": "https://registry.npmjs.org/braces/-/braces-2.3.2.tgz",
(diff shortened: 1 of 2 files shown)


kare:
@hpihkala I just saw this https://github.com/streamr-dev/streamr-platform/blob/development/app/package.json#L256 and felt enlightened

hpihkala:
🤯

hpihkala:
@timoxley could you try adding a similar note to `package.json` here that the `babel-runtime` needs to be in `dependencies`?

hpihkala:
Merged and published `2.2.1`. RunKit no longer shows an error.

timoxley:
@hpihkala yeah so you can't just kill the dev deps then run the tests since you need dev dependencies in order to run the tests. You want to run the tests against the built bundle without that bundle being able to pick up the devDependencies, and since it'll pick up anything in a parent dir's `node_modules` you'd need to copy the bundle into a directory outside the project dir e.g. temp and use the project's dev deps to run the tests. Perhaps not worth the hassle.

----

A simple check I've done in the past is to remove dev deps and check the dist file still parses:

```
npm run prune --production && node dist/index.js
```

Any missing dependencies should then crash the process. 
This could be run as a `prepublishOnly` lifecycle script.

PR #92, opened 2019-09-05
Add package.json note about babel-runtime vs devDependencies

timoxley:
As requested in https://github.com/streamr-dev/streamr-client-javascript/pull/91#issuecomment-527823682


hpihkala:
@timoxley thanks!

PR #93, opened 2019-09-10
Fix problem with CombinedSubscription swapping out HistoricalSubscription for RealtimeSubscription and confusing the secondResend

jtakalai:
also some defense in case someone tries to requestResend RealtimeSubscriptions without resendOptions

no idea how to test that so didn't write a test... probably like: mock connection, emit "resend done" and see if it explodes. Timing is kind of important too though, to make secondResend still happen.

Also not sure if in fact one of the other handlers should catch it, or if they ever get activated. A flowchart or something might help seeing what can and shouldn't happen at each state.

### DIFF
--- a/src/AbstractSubscription.js
+++ b/src/AbstractSubscription.js
@@ -97,7 +97,7 @@ export default class AbstractSubscription extends Subscription {
             try {
                 this.emit('no_resend', response)
             } finally {
-                this._finishResend()
+                this._finishResend(true)
             }
         })
     }
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -570,10 +570,12 @@ export default class StreamrClient extends EventEmitter {
                     // another resend if it doesn't. So we can anyway clear this resend request.
                     const handler = () => {
                         clearTimeout(secondResend)
+                        sub.removeListener('resend done', handler)
                         sub.removeListener('message received', handler)
                         sub.removeListener('unsubscribed', handler)
                         sub.removeListener('error', handler)
                     }
+                    sub.once('resend done', handler)
                     sub.once('message received', handler)
                     sub.once('unsubscribed', handler)
                     sub.once('error', handler)
@@ -630,8 +632,13 @@ export default class StreamrClient extends EventEmitter {
                 options.publisherId || null, options.msgChainId || null, sessionToken,
             )
         }
-        debug('_requestResend: %o', request)
-        this.connection.send(request)
+
+        if (request) {
+            debug('_requestResend: %o', request)
+            this.connection.send(request)
+        } else {
+            this.handleError("Can't _requestResend without resendOptions")
+        }
     }
 
     _requestPublish(streamMessage, sessionToken) {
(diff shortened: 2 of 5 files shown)


src/StreamrClient.js
```diff
@@ -570,10 +570,12 @@ export default class StreamrClient extends EventEmitter {
                     // another resend if it doesn't. So we can anyway clear this resend request.
```
jtakalai: will this work correctly now though? If messages haven't been stored yet, shouldn't the network say "resend done" and "no_resend"? Then we *should* do the secondResend, but also we should *not* swap HistoricalSubscription for RealtimeSubscription yet, I think? Only after second resend confirms there really was nothing to resend?

mthambipillai: I think you're right. The following scenario is valid:
1) subscribe with resend options
2) a first resend request is sent
3) the network answers with "no_resend" but maybe some messages are still being propagated and waiting to be stored
4) The HistoricalSubscription should NOT emit "resend done"
5) After some delay, the second resend request is sent
6) The network again answers with "no_resend". We assume the delay is large enough so that this response is correct.
7) This time the HistoricalSubscription can emit "resend done" and we switch to RealTimeSubscription

I added a test to reproduce your reported bug and the scenario above. I did a small fix in HistoricalSubscription to be able to differentiate between the first and the second "no_resend" answers from the network.

src/AbstractSubscription.js
```diff
@@ -97,7 +97,7 @@ export default class AbstractSubscription extends Subscription {
             try {
                 this.emit('no_resend', response)
```
jtakalai: First look at this code I was "wait what... how can emit throw?" but of course it can since it synchronously calls the handlers. This is bad if code is meant for library use. Consider:
```
const EE = require("events")

const ee = new EE()

ee.on("error", console.error)

ee.on("asdf", console.log)
ee.on("asdf", console.log)
ee.on("asdf", x => {
    throw new Error(x)
})
ee.on("asdf", console.log)
ee.on("asdf", console.log)

ee.emit("asdf", "testing lol")
```
prints
```
testing lol
testing lol
/Users/jtakalai/temp/eventEmitterTest.js:10
    throw new Error(x)
    ^

Error: testing lol
```
Meaning the order of handlers causes some be called and some not if a handler throws in the middle. That's just not cool. For our code, we can control that, but if "no_resend" is something someone else could `.on`, then we should use something like https://github.com/sindresorhus/emittery#isnt-eventemitter-synchronous-for-a-reason instead (and ofc then we can't write try catch for it, instead `emit` returns a Promise that can be `.catch`ed)

jtakalai: Should I make a ticket about switch Node EventEmitter for emittery (or similar async thing)? It could be prioritized low or whatever, but I think it's a consideration for a library.

mthambipillai: "no_resend" is meant for internal use only. But anyway, it seems to be another topic so I'll merge this fix if that's ok. Sure you can make a ticket, but maybe someone with more JS experience than me should give his opinion on this.

PR #94, opened 2019-09-17
CPS-35: Add Community Product functions

kare:


src/rest/CommunityEndpoints.js
```diff
@@ -0,0 +1,61 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+            },
+        },
+    )
+    return json
+}
+
+export async function withdraw(communityAddress, memberAddress, wallet) {
+    const json = await authFetch(
```
hpihkala: This call doesn't need to be authenticated? Looks like we could reuse the `memberStats` function here, right?

src/rest/CommunityEndpoints.js
```diff
@@ -0,0 +1,61 @@
...
+    const contract = new Contract(communityAddress, CommunityProduct.abi, wallet)
+    const withdrawTx = await contract.withdrawAll(json.withdrawableBlockNumber, json.withdrawableEarnings, json.proof)
+    await withdrawTx.wait(2)
+}
+
+export async function communityStats(communityAddress) {
+    const json = fetch(
+        `${this.options.restUrl}/communities/${communityAddress}/stats`,
+        {
+            method: 'GET',
```
hpihkala: `GET` is the default, could be omitted

src/rest/CommunityEndpoints.js
```diff
@@ -0,0 +1,61 @@
...
+        },
+    ).then((res) => res.json())
+    return json
+}
+
+export async function memberStats(communityAddress, memberAddress) {
+    const json = await fetch(
+        `${this.options.restUrl}/communities/${communityAddress}/members/${memberAddress}`,
+        {
+            method: 'GET',
```
hpihkala: `GET` is the default, could be omitted

src/rest/CommunityEndpoints.js
```diff
@@ -0,0 +1,61 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+            }
+        }
+    ).then((res) => res.json())
+
+    const contract = new Contract(communityAddress, CommunityProduct.abi, wallet)
+    const withdrawTx = await contract.withdrawAll(json.withdrawableBlockNumber, json.withdrawableEarnings, json.proof)
+    await withdrawTx.wait(2)
```
hpihkala: I'm guessing `2` is the number of block confirmations. We should probably allow the calling application to choose instead of hard-coding a value:
```
export async function withdraw(communityAddress, memberAddress, wallet, confirmations = 1) {
...
```

kare:
@hpihkala Thanks!

src/rest/CommunityEndpoints.js
```diff
@@ -23,39 +23,25 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
     return json
 }
 
-export async function withdraw(communityAddress, memberAddress, wallet) {
-    const json = await authFetch(
+export async function memberStats(communityAddress, memberAddress) {
+    const json = await fetch(
```
jtakalai: is this really eslint-recommended style? Remove the `{}` on put on one line maybe. And instead of await, just return the Promise.

src/rest/CommunityEndpoints.js
```diff
@@ -23,39 +23,25 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
-                'Content-Type': 'application/json',
-            }
-        }
+        {}
     ).then((res) => res.json())
+    return json
+}
 
+export async function withdraw(communityAddress, memberAddress, wallet, confirmations = 1) {
+    const stats = memberStats(communityAddress, memberAddress)
```
jtakalai: does this really work without await?

src/rest/CommunityEndpoints.js
```diff
@@ -23,39 +23,25 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
     const contract = new Contract(communityAddress, CommunityProduct.abi, wallet)
-    const withdrawTx = await contract.withdrawAll(json.withdrawableBlockNumber, json.withdrawableEarnings, json.proof)
-    await withdrawTx.wait(2)
+    const withdrawTx = await contract.withdrawAll(stats.withdrawableBlockNumber, stats.withdrawableEarnings, stats.proof)
+    await withdrawTx.wait(confirmations)
 }
 
 export async function communityStats(communityAddress) {
     const json = fetch(
```
jtakalai: same as `memberStats`

src/rest/CommunityEndpoints.js
jtakalai: Tests DO pass, but... eh.

src/rest/CommunityEndpoints.js
```diff
@@ -0,0 +1,47 @@
+import fetch from 'node-fetch'
+import { Contract } from 'ethers'
+
+import authFetch from './authFetch'
+
+const CommunityProduct = require('../../CommunityProduct.json')
+
+export async function joinCommunity(communityAddress, memberAddress, secret = undefined) {
```
jtakalai: Backend only allows the user to join who is logged in the session. So `memberAddress` is possibly redundant, it should be taken from the authenticated session if the session was initiated using private key (ethereum challenge response auth).

jtakalai: Starting to think about it... why does backend need it either (in the POST body) if it's supposed to only work for the authenticated address? Maybe you SHOULD be able to generate joinRequests for others, too?

jtakalai: I guess there's always the spamming problem, but ofc it doesn't cost anything to authenticate with StreamrClient right now, so in practice it's equivalent.

kare:
@mthambipillai Yes, we are working on tests with @jtakalai

PR #95, opened 2019-10-18
Add failing realtime + resend test.

timoxley:
I think this test captures at least one of the issues I was having with the client.

It seems resends work, and realtime works, but AFAICT they don't seem to work together.

Interestingly, this test passes when cherry-picked on top of `2.2.1`, so it's likely something to do with the [changes between `2.2.1` and `2.2.2`](https://github.com/streamr-dev/streamr-client-javascript/compare/e479b68927b8c6c5ce5d4fea8e8eb4309e55e9ec...7cb6790e2f53b984a2d8aa191ddda13f0df03b39#diff-b33004f77181ebaafb1f0c41f5c34f07R45-R53), likely this commit: 36058631d48fe702eb3e1cd4738fd5f06fe88e4b


hpihkala:
@mthambipillai could you take a look?

PR #96, opened 2019-10-23
Identify subscriptions by streamId+partition, not by streamId

hpihkala:
`SubscribedStream`s were indexed by `streamId` only, and subscriptions to different partitions didn't work. If a client wanted to subscribe to partitions 2, 3, and 4:

```
const handler = (payload, streamMessage) => {
    console.log(`Got message ${JSON.stringify(payload)} from partition ${streamMessage.getStreamPartition()}`)
}

[2,3,4].forEach(partition => {
    client.subscribe(
        {
            stream: 'YRXGkm7dSbWJ2H3c2EsiIw',
            partition: partition,
        },
        handler,
    )
})
```

The client would only send a `SubscribeRequest` for partition 2. The other two calls to `subscribe` would add the handler for partitions 3 and 4 to the existing `SubscribedStream` which was created for partition 2. Due to this, the client would only subscribe to partition 2 and the handler would be called 3 times for each message.

This PR fixes the issue by repurposing and renaming `SubscribedStream` to `SubscribedStreamPartition`, and indexing them by a combination of `streamId + streamPartition`.

Loose end: In `SubscribedStream` (now `SubscribedStreamPartition`), there is some caching of stream objects and stream publishers going on. It is redundant to request and cache these for each subscribed stream-partition, as these are stream-level things. However, I needed to get things working urgently, so I'm leaving the caching refactor out of scope for this PR, and creating a follow-up ticket about it.

### DIFF
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -28,7 +28,7 @@ import HistoricalSubscription from './HistoricalSubscription'
 import Connection from './Connection'
 import Session from './Session'
 import Signer from './Signer'
-import SubscribedStream from './SubscribedStreamPartition'
+import SubscribedStreamPartition from './SubscribedStreamPartition'
 import Stream from './rest/domain/Stream'
 import FailedToPublishError from './errors/FailedToPublishError'
 import MessageCreationUtil from './MessageCreationUtil'
@@ -283,7 +283,7 @@ export default class StreamrClient extends EventEmitter {
     _addSubscription(sub) {
         let sp = this._getSubscribedStreamPartition(sub.streamId, sub.streamPartition)
         if (!sp) {
-            sp = new SubscribedStream(this, sub.streamId, sub.streamPartition)
+            sp = new SubscribedStreamPartition(this, sub.streamId, sub.streamPartition)
             this._addSubscribedStreamPartition(sp)
         }
         sp.addSubscription(sub)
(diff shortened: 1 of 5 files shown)


src/StreamrClient.js
```diff
@@ -28,7 +28,7 @@ import HistoricalSubscription from './HistoricalSubscription'
 import Connection from './Connection'
 import Session from './Session'
 import Signer from './Signer'
-import SubscribedStream from './SubscribedStream'
+import SubscribedStream from './SubscribedStreamPartition'
```
mthambipillai: Maybe `import SubscribedStreamPartition from './SubscribedStreamPartition'` ?

src/StreamrClient.js
```diff
@@ -259,26 +259,60 @@ export default class StreamrClient extends EventEmitter {
...
+        const key = subscribedStreamPartition.streamId + subscribedStreamPartition.streamPartition
+        delete this.subscribedStreamPartitions[key]
+    }
+
     _addSubscription(sub) {
-        if (!this.subscribedStreams[sub.streamId]) {
-            this.subscribedStreams[sub.streamId] = new SubscribedStream(this, sub.streamId)
+        let sp = this._getSubscribedStreamPartition(sub.streamId, sub.streamPartition)
+        if (!sp) {
+            sp = new SubscribedStream(this, sub.streamId, sub.streamPartition)
```
mthambipillai: `new SubscribedStreamPartition(...)`

PR #102, opened 2019-11-11
Cps 35 community endpoints

jtakalai:
"Version 0.2" of community / data union methods in StreamrClient


hpihkala:
Needs a reviewer.

test/integration/CommunityEndpoints.test.js
```diff
@@ -0,0 +1,246 @@
...
+    let community
+
+    let testProvider
+    let adminClient
+    let adminWallet
+
+    beforeAll(async () => {
+        testProvider = new providers.JsonRpcProvider(config.ethereumServerUrl)
+        const network = await testProvider.getNetwork().catch((e) => {
+            throw new Error(`Connecting to Ethereum failed, config = ${JSON.stringify(config)}`, e)
```
timoxley: Unfortunately the second argument here does nothing. Should never suppress the original error message, can easily lead to debugging the wrong thing when original message if not suppressed would have revealed the problem immediately.

One option would be to change this into a log:

```js
console.error(`Connecting to Ethereum failed, config = ${JSON.stringify(config)}`)
throw e
```

README.md
```diff
@@ -283,16 +283,54 @@ resend: {
...
+#### Admin functions
+
+Name | Returns | Description
+---- | ------- | -----------
+deployCommunity() | Transaction | Deploy a new community product
+createSecret(communityAddress, secret[, name]) | Create a secret for a community product
+communityIsReady(address) | | Wait until community is operated
+addMembers(communityAddress, memberAddressList) | | Add members
+kick(communityAddress, memberAddressList) | | Kick members out from community
+```
```
timoxley: perhaps  use \`\`\`js for syntax-highlighted code snippets?
i.e.

```js
const community = await client.deployCommunity()
community.address           // already available before deployment
await community.deployed()  // waits until contract is deployed
await community.isReady()   // waits until community is operated
```

instead of

```
const community = await client.deployCommunity()
community.address           // already available before deployment
await community.deployed()  // waits until contract is deployed
await community.isReady()   // waits until community is operated
```
```

File: test/integration/CommunityEndpoints.test.js [hunk 3]
```diff
@@ -0,0 +1,246 @@
...
+        it('can add and remove members', async () => {
+            console.log('starting test')
+            await adminClient.communityIsReady(community.address, console.log)
+
+            await adminClient.addMembers(community.address, memberAddressList, testProvider)
+            await adminClient.hasJoined(community.address, memberAddressList[0])
+            const res = await adminClient.getCommunityStats(community.address)
+            assert.deepStrictEqual(res.memberCount, {
+                total: 3, active: 3, inactive: 0
+            })
```
timoxley: Going forward, where possible we should use [jest's `expect` API](https://jestjs.io/docs/en/expect) instead of `assert`.

test/integration/CommunityEndpoints.test.js
```diff
@@ -0,0 +1,246 @@
...
+                provider: testProvider
+            })
+            assert.strictEqual(tr2.logs[0].address, adminClient.options.tokenAddress)
+
+            const walletAfter = await opToken.balanceOf(memberWallet.address)
+            const diff = walletAfter.sub(walletBefore)
+            assert.strictEqual(diff.toString(), res3.withdrawableEarnings)
+        }, 600000)
+
+        // TODO: test withdrawTo, withdrawFor, getBalance
```
timoxley: Should these be done in this PR?

test/integration/CommunityEndpoints.test.js
```diff
@@ -0,0 +1,246 @@
...
+import StreamrClient from '../../src'
+import * as Token from '../../contracts/TestToken.json'
+
+import config from './config'
+
+function sleep(ms) {
+    return new Promise((resolve) => {
+        setTimeout(resolve, ms)
+    })
+}
```
timoxley: perhaps https://github.com/streamr-dev/streamr-test-utils#wait

test/integration/.eslintrc.js
```diff
@@ -0,0 +1,18 @@
...
+    it: "readonly",
+    beforeAll: "readonly",
+    beforeEach: "readonly",
+    afterAll: "readonly",
+    afterEach: "readonly",
+  },
+
+  // some style exceptions for stuff that's useful in tests
+  rules: {
+    "no-console": 0,          // extra logging, obviously
```
timoxley: I do have a mild preference for just removing the messages and just adding logs back in if/when they're needed for debugging, but if you're keen on leaving these in then IMO this should remain as-is in favour of using `debug` for the test-suite specific debug-level messages. 

Using `debug` with a suite-specific prefix would hide this by default e.g. consider what happens to the signal vs noise of these messages if other tests running in parallel also want to log something like `'starting beforeEach'`. `debug` is already a dependency.

test/integration/CommunityEndpoints.test.js
```diff
@@ -0,0 +1,246 @@
...
+        community = await adminClient.deployCommunity({
+            provider: testProvider,
+        })
+        console.log(`Going to deploy to ${community.address}`)
+        await community.deployed()
+        console.log(`Deployment done for ${community.address}`)
+        await community.isReady(2000, 200000)
+        console.log(`Community ${community.address} is ready to roll`)
+        await adminClient.createSecret(community.address, 'secret', 'CommunityEndpoints test secret')
+    }, 300000)
```
timoxley: is there a way we can get this to take less than 5 minutes? This is a very long time.

If not, I wonder if we should move such glacially slow tests into a different directory so everything else can be tested easily.

test/integration/CommunityEndpoints.test.js
```diff
@@ -0,0 +1,246 @@
...
+                total: 3, active: 3, inactive: 0
+            })
+
+            await adminClient.kick(community.address, memberAddressList.slice(1), testProvider)
+            await sleep(1000) // TODO: instead of sleeping, find a way to check server has registered the parting
+            const res2 = await adminClient.getCommunityStats(community.address)
+            assert.deepStrictEqual(res2.memberCount, {
+                total: 3, active: 1, inactive: 2
+            })
+        }, 300000)
```
timoxley: very long timeout

test/integration/CommunityEndpoints.test.js
jtakalai: Did the change but... is it just me or doesn't `await sleep(1000)` look more readable than `await wait(1000)`... even `await milliseconds(1000)` would be better IMO

test/integration/CommunityEndpoints.test.js
jtakalai: I made another ticket of trimming that down https://streamr.atlassian.net/browse/CPS-85

I don't think it actually will take 5 minutes, but Travis is slow sometimes...

test/integration/CommunityEndpoints.test.js
jtakalai: I made another ticket of trimming that down https://streamr.atlassian.net/browse/CPS-85

README.md
jtakalai: changed the rest, too

test/integration/CommunityEndpoints.test.js
jtakalai: good catch. There are libraries that nest errors, but I'm not going to introduce one here

test/integration/CommunityEndpoints.test.js
jtakalai: made another ticket for that https://streamr.atlassian.net/browse/CPS-85

Now I'd just very much like to get these into master

hpihkala:
@jtakalai @timoxley what's the status, can we get this merged?

timoxley:
Ok this seems ready to go. Seems to pass reliably now.

PR #104, opened 2019-11-14
use request id instead of sub id

mthambipillai:
WIP

This PR changes the way resend requests and responses are coordinated. They used to be linked with the subscription id (which lead to problems when multiple resends were happening at the same time within a same subscription but different message chains). Now they use specific request ids.

This PR is blocked by this other PR: https://github.com/streamr-dev/streamr-client-protocol-js/pull/23

### DIFF
--- a/src/AbstractSubscription.js
+++ b/src/AbstractSubscription.js
@@ -38,7 +38,7 @@ export default class AbstractSubscription extends Subscription {
         })
     }
 
-    addPendingResendRequestIds(requestId) {
+    addPendingResendRequestId(requestId) {
         this.pendingResendRequestIds[requestId] = true
     }
 
--- a/src/AbstractSubscription.js
+++ b/src/AbstractSubscription.js
@@ -13,6 +13,7 @@ export default class AbstractSubscription extends Subscription {
     constructor(streamId, streamPartition, callback, groupKeys, propagationTimeout, resendTimeout, orderMessages = true) {
         super(streamId, streamPartition, callback, groupKeys, propagationTimeout, resendTimeout)
         this.callback = callback
+        this.pendingResendRequestIds = {}
         this.orderingUtil = (orderMessages) ? new OrderingUtil(streamId, streamPartition, (orderedMessage) => {
             this._handleInOrder(orderedMessage)
         }, (from, to, publisherId, msgChainId) => {
@@ -35,6 +36,10 @@ export default class AbstractSubscription extends Subscription {
         this.on('error', () => {
             this._clearGaps()
         })
+    }
+
+    addPendingResendRequestIds(requestId) {
+        this.pendingResendRequestIds[requestId] = true
     }
 
     _handleInOrder(orderedMessage) {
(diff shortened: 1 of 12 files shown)


src/AbstractSubscription.js
```diff
@@ -37,6 +38,10 @@ export default class AbstractSubscription extends Subscription {
         })
     }
 
+    addPendingResendRequestIds(requestId) {
```
harbu: mv `addPendingResendRequestIds ` => `addPendingResendRequestId`

src/CombinedSubscription.js
```diff
@@ -47,6 +47,10 @@ export default class CombinedSubscription extends Subscription {
         return this.sub.stop()
     }
 
+    addPendingResendRequestIds(requestId) {
```
harbu: mv `addPendingResendRequestIds ` => `addPendingResendRequestId`

src/ResendUtil.js
```diff
@@ -0,0 +1,35 @@
...
+        this.counter = 0
+    }
+
+    generateRequestId() {
+        const id = this.counter
+        this.counter += 1
+        return id.toString()
+    }
+
+    getSubFromResendResponse(response, responseTypeName) {
```
harbu: It is generally expected that functions and methods with name `getXXX` do not modify state.

src/ResendUtil.js
```diff
@@ -0,0 +1,35 @@
...
+        if (!this.subForRequestId[response.subId]) { // TODO: replace with response.requestId
+            const error = new Error(`Received unexpected ${responseTypeName} message ${response.serialize()}`)
```
harbu: Could we use `response.constructor.name` to achieve the same and get rid of one extra parameter?

test/integration/StreamrClient.test.js
```diff
@@ -210,7 +210,7 @@ describe('StreamrClient Connection', () => {
                 ])
                 done()
             })
-        })
+        }, 300000)
```
harbu: This is quite a long timeout. We can probably declare the test failing in e.g. 30 seconds or 1 minute instead of 5 minutes.

test/unit/StreamrClient.test.js
```diff
@@ -437,18 +451,27 @@ describe('StreamrClient', () => {
...
 
             it('emits event on associated subscription', () => {
                 sub.handleResending = sinon.stub()
-                const resendResponse = ResendResponseResending.create(sub.streamId, sub.streamPartition, sub.id)
+                const resendResponse = ResendResponseResending.create(sub.streamId, sub.streamPartition, 0)
                 connection.emitMessage(resendResponse)
                 sinon.assert.calledWith(sub.handleResending, resendResponse)
             })
-            it('ignores messages for unknown subscriptions', () => {
+            it('throws when unknown request id', (done) => {
```
harbu: `emits error when unknown request id`?

test/unit/StreamrClient.test.js
```diff
@@ -458,18 +481,27 @@ describe('StreamrClient', () => {
...
+                        last: 5,
+                    },
+                })
+                connection.expect(ResendLastRequest.create('stream1', 0, 0, 5, 'session-token'))
             })
 
             it('calls event handler on subscription', () => {
                 sub.handleNoResend = sinon.stub()
-                const resendResponse = ResendResponseNoResend.create(sub.streamId, sub.streamPartition, sub.id)
+                const resendResponse = ResendResponseNoResend.create(sub.streamId, sub.streamPartition, 0)
```
harbu: We'll want to make these into strings for upcoming validation version.

PR #108, opened 2019-12-17
Fix subscribe with resend

mirpo:


mirpo:
I don't know why `second resend request` was added, but if you'll try run my tests with it, they will fail.

package.json
```diff
@@ -52,6 +52,7 @@
     "express": "^4.17.1",
     "jest": "^23.6.0",
     "sinon": "^7.3.2",
+    "streamr-test-utils": "^1.0.4",
```
timoxley: 👍

test/integration/Resends.test.js
```diff
@@ -0,0 +1,170 @@
...
+            })
+
+            sub.on('subscribed', async () => {
+                for (let i = MAX_MESSAGES; i < MAX_MESSAGES * 2; i++) {
+                    const message = {
+                        msg: `message${i}`,
+                    }
+
+                    // eslint-disable-next-line no-await-in-loop
+                    await client.publish(stream.id, message)
```
timoxley: maybe `client.publish(stream.id, message).then(() => publishedMessages.push(message))` so it publishes in parallel and the half-minute timeout could be lowered?

test/integration/Resends.test.js
```diff
@@ -0,0 +1,170 @@
...
+
+                // eslint-disable-next-line no-loop-func
+                sub.once('resent', () => {
+                    expect(receivedMessages).toStrictEqual(publishedMessages)
+                })
+
+                // eslint-disable-next-line no-await-in-loop
+                await waitForCondition(() => receivedMessages.length === MAX_MESSAGES, 10000)
+            }
+        }, 50000)
```
timoxley: oh man, not your fault but this is starting to get ridiculous

test/integration/Resends.test.js
```diff
@@ -0,0 +1,170 @@
...
+
+        it('resend last using subscribe function', async () => {
+            for (let i = 0; i < TEST_REPEATS; i++) {
```
timoxley: one alternative way of achieving this is to do something like this: 

https://github.com/streamr-dev/engine-and-editor/pull/707/files#diff-71db015fd49f36fd3008ea2fdb12148dR315-R395

i.e. put the `it` inside the `for` loop.

This has the benefit of telling you which repeat failed and while also allowing you to set a more normal timeout for each individual test, which should allow repeats to fail faster. e.g. in the example above the test would always pass on the first 2-3 times, but would fail sporadically after that. This was useful debugging information.

Just a suggestion.

test/integration/Resends.test.js
```diff
@@ -0,0 +1,170 @@
...
+        let publishedMessages
+
+        beforeEach(async () => {
+            client = createClient()
+            await client.ensureConnected()
+
+            publishedMessages = []
+
+            stream = await client.createStream({
+                name: uuid(),
```
timoxley: tip: doesn't make a lot of difference but consider using [`lodash/uniqueId`](https://lodash.com/docs/4.17.15#uniqueId) instead of `uuid`.

`uniqueId` generates a string using incrementing numbers and an optional custom prefix e.g. `stream1`, `stream2`.

This can help a lot with debugging async behaviour as the incrementing numbers clearly show what order things occurred in so you can more clearly see misorderings or even cases where data is leaking across tests. e.g. `stream10` arrives in a test where the first stream should have been `stream11`.

The downside is that unlike `uuid` it's not guaranteed to be unique across processes, so when Jest runs suites in parallel, you can end up with more than one `stream1`, but I've found confusion due to this to be rare and the benefits of human readable/sortable uniqueness outweighs the possible downsides. I don't think a universally unique name is necessary here, but if you do want more cross-process uniqueness you can do something like:

```JS
const uid = (prefix) => uniqueId(`p${process.pid}-${prefix}`)
```

test/integration/Resends.test.js
```diff
@@ -0,0 +1,170 @@
...
+                    last: MAX_MESSAGES,
+                },
+            }, (message) => {
+                receivedMessages.push(message)
+            })
+
+            sub.on('subscribed', async () => {
+                for (let i = MAX_MESSAGES; i < MAX_MESSAGES * 2; i++) {
+                    const message = {
+                        msg: `message${i}`,
```
timoxley: `lodash/uniqueId` could be useful here too.

src/HistoricalSubscription.js
```diff
@@ -42,14 +42,8 @@ export default class HistoricalSubscription extends AbstractSubscription {
...
-    finishResend(isNoResend = false) {
-        // if after a first resend request no messages are received (received ResendResponseNoResend),
-        // we wait for the response to the second resend request before considering the resend done (messages might have been stored in between)
-        if (isNoResend && !this.firstNoResendReceived) {
-            this.firstNoResendReceived = true
-        } else {
-            this._lastMessageHandlerPromise = null
-            this.emit('resend done')
-        }
+    finishResend() {
```
timoxley: I think this means `finishResend(true)` can be removed from https://github.com/streamr-dev/streamr-client-javascript/blob/93b49f90872186ac31246cc2a0cd46ec6a973c98/src/AbstractSubscription.js#L107

test/integration/Resends.test.js
mirpo: Good point 👍

mirpo:
> Do we need it to retry the resend request?
> Left some non-critical suggestions, but otherwise lgtm 👍

Current implementation (`second resend`) just gives us random errors. So if we need `retry`, I'd like to have as a separate ticket for it. And we can start with proper integration test of `retry` instead of mocking everything.

test/integration/Resends.test.js
mirpo: I tried, but I need order :) Maybe it's another issue...

package.json
```diff
@@ -66,6 +66,7 @@
     "ethers": "^4.0.41",
     "eventemitter3": "^4.0.0",
     "heap": "^0.2.6",
+    "lodash.uniqueid": "^4.0.1",
```
timoxley: No need to do this any more, now you can depend on `lodash` directly and then do:

```js
import uniqueId from 'lodash/uniqueid'
```

 What you've got is fine for now, but there are many useful, time/code saving features in `lodash` so IMO better to just pull it in as a full dep so it's all readily available.

test/integration/Resends.test.js
```diff
@@ -0,0 +1,173 @@
...
+                })
+
+                // eslint-disable-next-line no-await-in-loop
+                await waitForCondition(() => receivedMessages.length === MAX_MESSAGES, 10000)
+            }, 10000)
+        }
+
+        for (let i = 0; i < TEST_REPEATS; i++) {
+            // eslint-disable-next-line no-loop-func
+            it(`resend last using subscribe function on try ${i}`, async () => {
```
timoxley: 👍

test/utils.js
```diff
@@ -0,0 +1,3 @@
+const uniqueId = require('lodash.uniqueid')
+
+export const uid = (prefix) => uniqueId(`p${process.pid}${prefix ? '-' + prefix : ''}`)
```
timoxley: 👍

PR #109, opened 2019-12-18
Handle the case where $http_code is not set

jtakalai:
this seems to cause lots of "/home/travis/.travis/functions: line 109: [: =: unary operator expected" messages in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ jobs:
       - git clone https://github.com/streamr-dev/streamr-docker-dev.git
       - sudo ifconfig docker0 10.200.10.1/24
       - "$TRAVIS_BUILD_DIR/streamr-docker-dev/streamr-docker-dev/bin.sh start 5"
-      - while true; do http_code=$(curl -s -o /dev/null -w "%{http_code}" http://localhost/api/v1/users/me); if [ $http_code = 401 ]; then echo "EE up and running"; break; else echo "EE not receiving connections"; sleep 5s; fi; done
-      - while true; do http_code=$(curl -s -o /dev/null -w "%{http_code}" http://localhost/api/v1/volume); if [ $http_code = 200 ]; then echo "brokers up and running"; break; else echo "brokers not receiving connections"; sleep 5s; fi; done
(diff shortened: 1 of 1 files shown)


kare:
@jtakalai You'll need to quote the variable:
`if [ $http_code = 401 ]` -> `if [ "$http_code" = 401 ]`
and use equality operator instead of assignment:
`if [ "$http_code" = 401 ]` -> `if [ "$http_code" == "401" ]`

.travis.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       - git clone https://github.com/streamr-dev/streamr-docker-dev.git
       - sudo ifconfig docker0 10.200.10.1/24
       - "$TRAVIS_BUILD_DIR/streamr-docker-dev/streamr-docker-dev/bin.sh start 5"
-      - while true; do http_code=$(curl -s -o /dev/null -w "%{http_code}" http://localhost/api/v1/users/me); if [ $http_code = 401 ]; then echo "EE up and running"; break; else echo "EE not receiving connections"; sleep 5s; fi; done
-      - while true; do http_code=$(curl -s -o /dev/null -w "%{http_code}" http://localhost/api/v1/volume); if [ $http_code = 200 ]; then echo "brokers up and running"; break; else echo "brokers not receiving connections"; sleep 5s; fi; done
+      - while true; do http_code=$(curl -s -o /dev/null -w "%{http_code}" http://localhost/api/v1/users/me); if [ -z $http_code ] || [ $http_code = 401 ]; then echo "EE up and running"; break; else echo "EE not receiving connections"; sleep 5s; fi; done
```
timoxley: A `401` indicates the server is up, but IIUC matching `-z` means there was no http response code… does no response code mean the server is up?

I'm thinking the test logic here is backwards? i.e. Should it be `-n` and `&&` instead of `-z` and `||`? 

I think I've encountered this exact thing before myself when doing this check for platform, curious under what conditions can it respond with no response code? It would seem that could only happen if the connection was closed before a valid http response code was sent?

timoxley: For `platform` we use `-eq` instead of `==` https://github.com/streamr-dev/streamr-platform/blob/6d368300108d54cab707a03b029b7e588a10ea25/app/travis_scripts/utils.sh#L11

timoxley: Actually, now I look at it the correct answer is probably to check the [exit code of the curl command](https://linux.die.net/man/1/curl):

so something like:

`if [ $? -eq 0 ] && [ "$http_code" == "401" ]`

jtakalai:
I see this has already been implemented elsewhere (github actions)

PR #110, opened 2019-12-24
Disable Regenerator

timoxley:
Builds on top of #107.

In trying to figure out what's going on with the assertion order change mentioned in #107, I disabled `regenerator` so I could see the stack trace correctly and voila once it was using real async/await the original assertion order worked.

The `preset-env` `"last two versions"` config includes IE11 (or some other browser) which requires transpiling `async`/`await` with `regenerator`. Copied/adapted our browser support list & `core-js`/`runtime`/`preset-env` config [across from streamr-platform](https://github.com/streamr-dev/streamr-platform/blob/6d368300108d54cab707a03b029b7e588a10ea25/app/babel.config.js). 

This should make debugging overall easier since you will be able to see real async stack traces now.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,23 +1,34 @@
 {
-  "presets": [
-    [
-      "@babel/preset-env",
-      {
-        "modules": false,
-        "loose": false,
-        "targets": {
-          "browsers": [
-            "last 2 versions"]
-        }
-      }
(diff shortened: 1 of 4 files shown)


mirpo:
So I can copy later the same babel config to the protocol?

PR #111, opened 2020-01-02
Config basic actions

mirpo:
Migrated from travis to github actions.
- unit tests
- integrations tests
- eslint
- node 10 + 12
- publish to npm (need to run once, but I tested similar config in js-protocol)


.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,59 @@
+name: Publish
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build:
+    name: Test on node 10 and ubuntu-latest
```
harbu: I wonder if there is a way to re-use the  jobs/tasks from `nodejs.yml` to make maintaining easier.

PR #117, opened 2020-01-22
Community products related features

jtakalai:
also README update reflecting those changes

Note that these changes fully reflect core review suggestions given for these files (well, README.md) in https://github.com/streamr-dev/streamr-client-javascript/pull/102 (thanks Tim!)

Tests are still found in that pull request, in their current state.

Why should this PR be accepted: it improves what there currently is in master. Current code in master doesn't have tests, and it was written long time ago. Any way you look at it, the code given here is more tested and more polished. It contains features that, had the code been available earlier to front-end, might have prevented code (and work) duplication.

### DIFF
--- a/src/rest/CommunityEndpoints.js
+++ b/src/rest/CommunityEndpoints.js
@@ -119,18 +119,13 @@ function parseWalletFromOptions(client, options) {
  */
 export async function deployCommunity(options) {
     const wallet = parseWalletFromOptions(this, options)
-
-    const blockFreezePeriodSeconds = options.blockFreezePeriodSeconds || 0
-    const adminFee = options.adminFee || 0
-    const tokenAddress = options.tokenAddress || this.options.tokenAddress
-    const streamrNodeAddress = options.streamrNodeAddress || this.options.streamrNodeAddress
-    const streamrOperatorAddress = options.streamrOperatorAddress || this.options.streamrOperatorAddress
-    // const {blockFreezePeriodSeconds, adminFeeFraction, tokenAddress, streamrNodeAddress, operatorAddress} = {
-    //     blockFreezePeriodSeconds: 0,
-    //     adminFee: 0,
-    //     ...this.options,
-    //     ...options,
-    // }
+    const {
+        blockFreezePeriodSeconds = 0,
+        adminFee = 0,
+        tokenAddress = this.options.tokenAddress,
+        streamrNodeAddress = this.options.streamrNodeAddress,
+        streamrOperatorAddress = this.options.streamrOperatorAddress
+    } = options
 
     await throwIfNotContract(wallet.provider, tokenAddress, 'options.tokenAddress')
     await throwIfBadAddress(streamrNodeAddress, 'options.streamrNodeAddress')
--- a/src/rest/CommunityEndpoints.js
+++ b/src/rest/CommunityEndpoints.js
@@ -35,7 +35,7 @@ function throwIfBadAddress(address, variableDescription) {
     try {
         return getAddress(address)
     } catch (e) {
-        throw new Error(`${variableDescription || 'Error'}: Bad Ethereum address ${address}`)
+        throw new Error(`${variableDescription || 'Error'}: Bad Ethereum address ${address}. Original error: ${e.stack}.`)
     }
 }
 
@@ -163,20 +163,19 @@ export async function deployCommunity(options) {
  * Await this function when you want to make sure a community is deployed and ready to use
  * @param {EthereumAddress} address of the community
  * @param {Number} pollingIntervalMs (optional, default: 1000) ask server if community is ready
- * @param {Number} timeoutMs (optional, default: 60000) give up
+ * @param {Number} retryTimeoutMs (optional, default: 60000) give up sending more retries
  * @return {Promise} resolves when community server is ready to operate the community (or fails with HTTP error)
  */
-export async function communityIsReady(address, pollingIntervalMs, timeoutMs) {
+export async function communityIsReady(address, pollingIntervalMs = 1000, retryTimeoutMs = 60000) {
     let stats = await get(this, address, '/stats')
-    const timeout = timeoutMs || 60000
     const startTime = Date.now()
-    while (stats.error && Date.now() < startTime + timeout) {
+    while (stats.error && Date.now() < startTime + retryTimeoutMs) {
         debug(`Waiting for community ${address} to start. Status: ${JSON.stringify(stats)}`)
-        await sleep(pollingIntervalMs || 1000) // eslint-disable-line no-await-in-loop
+        await sleep(pollingIntervalMs) // eslint-disable-line no-await-in-loop
         stats = await get(this, address, '/stats') // eslint-disable-line no-await-in-loop
     }
     if (stats.error) {
-        throw new Error(`Community failed to start within ${timeout} ms. Status: ${JSON.stringify(stats)}`)
+        throw new Error(`Community failed to start, retried for ${retryTimeoutMs} ms. Status: ${JSON.stringify(stats)}`)
     }
 }
 
@@ -243,10 +242,10 @@ export async function joinCommunity(communityAddress, secret) {
  * @param {EthereumAddress} communityAddress
  * @param {EthereumAddress} memberAddress (optional, default is StreamrClient's auth: privateKey)
  * @param {Number} pollingIntervalMs (optional, default: 1000) ask server if member is in
- * @param {Number} timeoutMs (optional, default: 60000) give up
+ * @param {Number} retryTimeoutMs (optional, default: 60000) give up
  * @return {Promise} resolves when member is in the community (or fails with HTTP error)
  */
-export async function hasJoined(communityAddress, memberAddress, pollingIntervalMs, timeoutMs) {
+export async function hasJoined(communityAddress, memberAddress, pollingIntervalMs = 1000, retryTimeoutMs = 60000) {
     let address = memberAddress
     if (!address) {
         const authKey = this.options.auth && this.options.auth.privateKey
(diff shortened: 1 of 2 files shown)


src/rest/CommunityEndpoints.js
```diff
@@ -1,24 +1,200 @@
...
+    const adminFeeBN = new BigNumber((adminFee * 1e18).toFixed()) // last 2...3 decimals are going to be gibberish
+
+    const stream = await this.createStream({
+        name: `Join-Part-${wallet.address.slice(0, 10)}-${Date.now()}`
+    })
+    debug(`Stream created: ${JSON.stringify(stream.toObject())}`)
+    const res1 = await stream.grantPermission('read', null)
+    debug(`Grant read permission response from server: ${JSON.stringify(res1)}`)
+    const res2 = await stream.grantPermission('write', streamrNodeAddress)
+    debug(`Grant write permission response to ${streamrNodeAddress} from server: ${JSON.stringify(res2)}`)
```
tumppi: Could we maybe extract this to a separate function? It would help us to unify `joinPartStream` creation on frontend since we cannot use `deployCommunity` directly because we need to watch the transaction.

src/rest/CommunityEndpoints.js
```diff
@@ -28,87 +204,286 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
-export async function withdraw(communityAddress, memberAddress, wallet, confirmations = 1) {
-    const stats = await this.memberStats(communityAddress, memberAddress)
-    if (!stats.withdrawableBlockNumber) {
-        throw new Error('No earnings to withdraw.')
+/**
+ * Send a joinRequest, or get into community instantly with a community secret
+ * @param {EthereumAddress} communityAddress to join
+ * @param {String} secret (optional) if given, and correct, join the community immediately
+ */
+export async function joinCommunity(communityAddress, secret) {
```
tumppi: I think we should add `partCommunity` or similar so that members themselves can leave the community.

src/rest/CommunityEndpoints.js
jtakalai: yes, although this `deployCommunity` works a bit differently from the old one: this one returns immediately with the ethers.js Transaction object (you can get the tx hash and the future contract address), and you can then either `await contract.deployed()` or even `await contract.isReady()` which means server has noticed the creation and started operating, so it's ready to use.

jtakalai: but you may be right about how it works together with UI stuff like Metamask popup. I imagine Metamask popup would appear at `deployer.deploy` call which creates the tx. But the function would return right after user gives the OK, and waiting for the tx would then be optional.

jtakalai: the issue with separating those is that they still need to be done together both or none, so if separation is needed, then probably better just duplicate the code, almost... depends on the case for sure.

src/rest/CommunityEndpoints.js
jtakalai: good idea actually! Only I'm not sure if it's even implemented in EE currently (probably not), so... it would have to be done by admin anyway, for which there is `kick`.

src/rest/CommunityEndpoints.js
tumppi: We are using `web3` on the frontend side currently so that Ethers object might not help.

We have this code currently duplicated on the frontend side and it's not that bad. The name we use for the stream is a bit different on our side.

src/rest/CommunityEndpoints.js
```diff
@@ -1,24 +1,200 @@
...
-    const url = `${this.options.restUrl}/communities/${communityAddress}/joinRequests`
+const debug = debugFactory('StreamrClient::CommunityEndpoints')
+
+/** @typedef {String} EthereumAddress */
+
+function throwIfBadAddress(address, variableDescription) {
+    try {
+        return getAddress(address)
+    } catch (e) {
+        throw new Error(`${variableDescription || 'Error'}: Bad Ethereum address ${address}`)
```
timoxley: if possible, should always check the signature of the error is what you expect before rethrowing.

The risk is that this reports "Bad Ethereum address" when the problem is something else entirely e.g. a `ReferenceError` and we've hidden the actual problem + stack trace and also misdirected the user to a non-problem.

src/rest/CommunityEndpoints.js
```diff
@@ -1,24 +1,200 @@
...
+    const timeout = timeoutMs || 60000
+    const startTime = Date.now()
+    while (stats.error && Date.now() < startTime + timeout) {
+        debug(`Waiting for community ${address} to start. Status: ${JSON.stringify(stats)}`)
+        await sleep(pollingIntervalMs || 1000) // eslint-disable-line no-await-in-loop
+        stats = await get(this, address, '/stats') // eslint-disable-line no-await-in-loop
+    }
+    if (stats.error) {
+        throw new Error(`Community failed to start within ${timeout} ms. Status: ${JSON.stringify(stats)}`)
+    }
```
timoxley: hm, so if the timeout is set to 1000ms but the `get` call takes 10000ms+, it won't cause the function to fail or short circuit. Maybe the parameter should be called `retryTimeoutMs` so it can't be interpreted as an actual max timeout, or perhaps the function should use something like `Promise.race` to implement a hard timeout limit:

```JS
await Promise.race([
  sleep(timeoutMs),
  retry(),
])
```

src/rest/CommunityEndpoints.js
```diff
@@ -28,87 +204,286 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
-        throw new Error(`${variableDescription || 'Error'}: Bad Ethereum address ${address}`)
+/**
+ * Await this function when you want to make sure a member is accepted in the community
+ * @param {EthereumAddress} communityAddress
+ * @param {EthereumAddress} memberAddress (optional, default is StreamrClient's auth: privateKey)
+ * @param {Number} pollingIntervalMs (optional, default: 1000) ask server if member is in
+ * @param {Number} timeoutMs (optional, default: 60000) give up
+ * @return {Promise} resolves when member is in the community (or fails with HTTP error)
+ */
+export async function hasJoined(communityAddress, memberAddress, pollingIntervalMs, timeoutMs) {
```
timoxley: I feel like these should all be named parameters rather than positional i.e. `export async function hasJoined({communityAddress, memberAddress, pollingIntervalMs, timeoutMs}) {`

timoxley: also whether these are positional or not, should probably set the defaults in the parameter definition rather than at usage location i.e. `export async function hasJoined(communityAddress, memberAddress, pollingIntervalMs = 1000, timeoutMs = 60000) {`

src/rest/CommunityEndpoints.js
```diff
@@ -28,87 +204,286 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
+        if (!authKey) {
+            throw new Error("StreamrClient wasn't authenticated with privateKey, and memberAddress argument not supplied")
+        }
+        address = computeAddress(authKey)
+    }
 
-    const stream = await this.getOrCreateStream({
-        name: `Join-Part-${wallet.address.slice(0, 10)}-${Date.now()}`
+    const stats = await get(this, communityAddress, `/members/${address}`)
+    if (stats.error || stats.earnings === '0') {
```
timoxley: Note there's a suppressed error here that you might want to expose on the returned object somehow, especially if `stats.error` isn't an "expected" error. Otherwise user can't differentiate actual zero balances from infrastructure failure.

src/rest/CommunityEndpoints.js
```diff
@@ -28,87 +204,286 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
+/**
+ * Get the tx promise for withdrawing all your earnings
+ * @param {EthereumAddress} communityAddress
+ * @param {EthereumOptions} options
+ * @returns {Promise<providers.TransactionResponse>} await on call .wait to actually send the tx
+ */
+export async function getWithdrawTx(communityAddress, options) {
+    const wallet = parseWalletFromOptions(this, options)
+    const stats = await this.getMemberStats(communityAddress, wallet.address) // throws on connection errors
+    if (!stats.withdrawableBlockNumber) {
```
timoxley: what if there's some unexpected `stats.error`?

src/rest/CommunityEndpoints.js
```diff
@@ -28,87 +204,286 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
+
+/**
+ * Get the tx promise for withdrawing all earnings on behalf of another member
+ * @param {EthereumAddress} communityAddress
+ * @param {EthereumOptions} options
+ * @returns {Promise<providers.TransactionResponse>} await on call .wait to actually send the tx
+ */
+export async function getWithdrawTxFor(memberAddress, communityAddress, options) {
+    const stats = await this.getMemberStats(communityAddress, memberAddress) // throws on connection errors
+    if (!stats.withdrawableBlockNumber) {
```
timoxley: check for `stats.error`?

src/rest/CommunityEndpoints.js
```diff
@@ -28,87 +204,286 @@ export async function joinCommunity(communityAddress, memberAddress, secret = un
...
+ * Withdraw earnings and "donate" them to the given address
+ * @param {EthereumAddress} communityAddress
+ * @param {EthereumAddress} recipientAddress the other member who gets its tokens out of the Community
+ * @param {EthereumOptions} options
+ * @returns {Promise<providers.TransactionResponse>} await on call .wait to actually send the tx
+ */
+export async function getWithdrawTxTo(recipientAddress, communityAddress, options) {
+    const wallet = parseWalletFromOptions(this, options)
+    const stats = await this.getMemberStats(communityAddress, wallet.address) // throws on connection errors
+    if (!stats.withdrawableBlockNumber) {
```
timoxley: `stats.error`?

src/rest/CommunityEndpoints.js
```diff
@@ -1,24 +1,200 @@
...
+ * @return {Promise<Contract>} has methods that can be awaited: contract is deployed (`.deployed()`), operator is started (`.isReady()`)
+ */
+export async function deployCommunity(options) {
+    const wallet = parseWalletFromOptions(this, options)
+
+    const blockFreezePeriodSeconds = options.blockFreezePeriodSeconds || 0
+    const adminFee = options.adminFee || 0
+    const tokenAddress = options.tokenAddress || this.options.tokenAddress
+    const streamrNodeAddress = options.streamrNodeAddress || this.options.streamrNodeAddress
+    const streamrOperatorAddress = options.streamrOperatorAddress || this.options.streamrOperatorAddress
```
timoxley: could use JS destructuring defaults syntax like:

```js
const {
  blockFreezePeriodSeconds = 0,
  adminFee = 0,
  tokenAddress = this.options.tokenAddress,
  // etc
} = options
```

src/rest/CommunityEndpoints.js
```diff
@@ -1,24 +1,200 @@
...
+    // const {blockFreezePeriodSeconds, adminFeeFraction, tokenAddress, streamrNodeAddress, operatorAddress} = {
+    //     blockFreezePeriodSeconds: 0,
+    //     adminFee: 0,
+    //     ...this.options,
+    //     ...options,
+    // }
```
timoxley: yeah, something like this

src/rest/CommunityEndpoints.js
jtakalai: True, but that one really is a convenience function. It says zero in case of failure; this may be spurious, true, but the "typical" stats.error should be member or community doesn't exist. Detecting infrastructure failure would be a good separate helper function maybe. Later, I won't start guessing right now what typical infra failure would look like.

PR #119, opened 2020-02-05
finish resend only when there are no more pending requests

mthambipillai:


README.md
```diff
@@ -281,6 +281,14 @@ resend: {
 }
 ```
 
+If you choose one of the above resend options when subscribing, you can listen on the completion of this resend by doing the following:
+
+```javascript
+client.on('initial_resend_done', () => {
```
hpihkala: @mthambipillai This should be subscription-specific? So `sub.on('initial_resend_done', ...)`?

Or would we trigger it on the `client` when *all* initial resends are done? That would make sense, but that's not what the current implementation does.

PR #121, opened 2020-02-06
fix subscribe on reconnect after resend only

mthambipillai:
Prevents resend() to send a subscribe request on reconnection. Currently it does nothing else (it does not send the resend request again).

### DIFF
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -403,7 +403,7 @@ export default class StreamrClient extends EventEmitter {
         // TODO remove _addSubscription after uncoupling Subscription and Resend
         sub.setState(Subscription.State.subscribed)
         this._addSubscription(sub)
-        sub.on('initial_resend_done', () => this._removeSubscription(sub))
+        sub.once('initial_resend_done', () => this._removeSubscription(sub))
         await this._requestResend(sub)
         return sub
     }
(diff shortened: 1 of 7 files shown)


test/unit/StreamrClient.test.js
```diff
@@ -586,6 +589,48 @@ describe('StreamrClient', () => {
...
+                }
+            }, () => {})
+            const msg1 = msg(sub.streamId, {}, '0')
+            connection.emitMessage(msg1)
+            const resendResponse = ResendResponseResent.create(sub.streamId, sub.streamPartition, '0')
+            connection.emitMessage(resendResponse)
+            await client.pause()
+            await client.connect()
+        })
+    })
```
hpihkala: A test case is missing perhaps: what happens if the disconnection happens during resend?

- Resend request is sent
- (`ResendResponseResent` is not received)
- disconnect
- reconnect

So what should happen? Well, ideally the resend request might get repeated on reconnect, but that's not super important, it could be handled on application level. Let's leave that out of scope. 

But I guess the client and sub should at least end up in a healthy state. Compared to the situation where the `ResendResponseResent` **is** received, now the `initial_resend_done` event is never fired, and the `sub` is never cleaned up via `this._removeSubscription(sub)`? Not sure if that matters much, I guess it's just a minor memory leak.

src/StreamrClient.js
```diff
@@ -401,9 +401,10 @@ export default class StreamrClient extends EventEmitter {
             this.options.subscriberGroupKeys[options.stream], this.options.gapFillTimeout, this.options.retryResendAfter, this.options.orderMessages)
 
         // TODO remove _addSubscription after uncoupling Subscription and Resend
+        sub.setState(Subscription.State.subscribed)
         this._addSubscription(sub)
+        sub.on('initial_resend_done', () => this._removeSubscription(sub))
```
timoxley: Should this be `.once`?

PR #125, opened 2020-03-04
Add failing tests for concurrent resend + subscribe.

timoxley:
When issuing both resend & realtime subscription requests for the same stream, the realtime subscription is never created.

The tests in this PR pass against streamr-client 3.1.2, but not against 3.1.3. 

[Diff between 3.1.2 & 3.13](https://github.com/streamr-dev/streamr-client-javascript/compare/bddbea18...c43591d)

Concrete problem occurs in the data union server which issues concurrent resend + subscribe requests for the same stream: https://github.com/streamr-dev/streamr-community-products/blob/640dfa45f4f520ea40602052101f85e295a50528/src/streamrChannel.js#L110-L127

```JS
        log(`Starting playback of ${this.stream.id}`)
        const playbackSub = await this.client.resend({
            stream: this.stream.id,
            resend: {
                from: {
                    timestamp: syncStartTimestamp || 1,
                    sequenceNumber: 0,
                },
            },
        }, emitMessage)

        const queue = []
        this.client.subscribe({
            stream: this.stream.id
        }, (msg, meta) => {
            const len = queue.push({msg, meta})
            log(`Got message ${JSON.stringify(msg)}, queue length = ${len}}`)
        })
```

The data union server perhaps could achieve the desired result without this, however I believe the client is definitely misbehaving here.

Also interestingly, this doesn't *seem* to be causing an issue with the production data union server, which is using client 3.1.3. It's possible the async timings in production somehow work around the issue, though it's not proven this issue is definitely never happening in production, perhaps it's intermittent.

----

When the `.subscribe` occurs after the `.resend`, `client._requestSubscribe` erroneously considers the resend subscription a duplicate subscription, and "insta-subscribes" to it:
https://github.com/streamr-dev/streamr-client-javascript/blob/c43591d617a0f167d82d44bee477a780dac3936e/src/StreamrClient.js#L648-L668

and then immediately unsubscribes the realtime subscription as soon as the resend is finished:
https://github.com/streamr-dev/streamr-client-javascript/blob/f652138d47d642746a840ef715cc61da59c8bb02/src/StreamrClient.js#L406

----

Issue appears to be triggered by this change in `3.1.3` that immediately adds `sub.setState(Subscription.State.subscribed)` to the historical subscription:
https://github.com/streamr-dev/streamr-client-javascript/commit/f652138d47d642746a840ef715cc61da59c8bb02#diff-490b0cafe3e660b3fba36db4786f005aR404

However I think the bug is actually in the way `_requestSubscribe` detects duplicate subscriptions, which in turn is really a shortcoming in how subscriptions are keyed into `client.subscribedStreamPartitions` i.e. subscriptions with the same id+partition but entirely different resend options are considered the same subscription.

https://github.com/streamr-dev/streamr-client-javascript/blob/c43591d617a0f167d82d44bee477a780dac3936e/src/StreamrClient.js#L275-L294

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "streamr-client",
-  "version": "3.1.2",
+  "version": "3.1.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -893,9 +893,9 @@
       }
     },
     "@babel/runtime-corejs3": {
-      "version": "7.7.7",
-      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs3/-/runtime-corejs3-7.7.7.tgz",
-      "integrity": "sha512-kr3W3Fw8mB/CTru2M5zIRQZZgC/9zOxNSoJ/tVCzjPt3H1/p5uuGbz6WwmaQy/TLQcW31rUhUUWKY28sXFRelA==",
+      "version": "7.8.4",
+      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs3/-/runtime-corejs3-7.8.4.tgz",
+      "integrity": "sha512-+wpLqy5+fbQhvbllvlJEVRIpYj+COUWnnsm+I4jZlA8Lo7/MJmBhGTCHyk1/RWfOqBRJ2MbadddG6QltTKTlrg==",
       "requires": {
         "core-js-pure": "^3.0.0",
         "regenerator-runtime": "^0.13.2"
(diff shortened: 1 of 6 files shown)


mthambipillai:
This is a result of mixing explicit resends with subscriptions, they should be uncoupled. @mirpo  you already stated that it should be uncoupled when you created the resend function: https://github.com/streamr-dev/streamr-client-javascript/blob/master/src/StreamrClient.js#L403. Did you have something in mind on how to do it? A quick solution would be to have 2 objects like [this one](https://github.com/streamr-dev/streamr-client-javascript/blob/master/src/StreamrClient.js#L67) instead of just 1: one for "explicit resends", one for subscriptions. But there might be something better.

Besides this bug, I think that

> The data union server perhaps could achieve the desired result without this

 is an understatement. It seems to me that [this whole code](https://github.com/streamr-dev/streamr-community-products/blob/640dfa45f4f520ea40602052101f85e295a50528/src/streamrChannel.js#L110-L145) could be made very simple using the intended feature:
```
this.client.subscribe({
  stream: this.stream.id
  resend: {
    from: {
      timestamp: syncStartTimestamp || 1,
      sequenceNumber: 0,
    },
  }
}, emitMessage)
```
But more importantly, the current code isn't correct and does not guarantee the desired result as it does not fill potential gaps between the end of the "playback" and the first real time message. But this is done under the hood by the `StreamrClient`, so its intented feature should be used. The explicit resend should be used for cases requiring historical messages only.

PR #126, opened 2020-03-11
Ignore case of userId when checking permissions.

timoxley:
Ethers-generated address are in checkum-case (mixed case) https://docs.ethers.io/ethers.js/html/notes.html#checksum-address, but the backend seems to report addresses in lowercase format only.

This will break any `userId === address`  checks that don't lowercase the userId beforehand. 
CI for any existing PRs will be failing on this issue.

I changed the two places where I saw `userId` being used, there may be others with a different variable name. 

I haven't dug into why this issue has only just appeared, but perhaps it's related to the permissions changes @kare has been working on?

### DIFF
--- a/src/rest/domain/Stream.js
+++ b/src/rest/domain/Stream.js
@@ -55,7 +55,7 @@ export default class Stream {
             if (userId === undefined) {
                 return !!p.anonymous // match nullish userId against p.anonymous
             }
-            return p.user.toLowerCase() === userIdCaseInsensitive // match against userId
+            return p.user && p.user.toLowerCase() === userIdCaseInsensitive // match against userId
         })
     }
 
--- a/src/rest/domain/Stream.js
+++ b/src/rest/domain/Stream.js
@@ -47,16 +47,25 @@ export default class Stream {
 
     async hasPermission(operation, userId) {
         const permissions = await this.getPermissions()
-        return permissions.find((p) => p.operation === operation && ((userId == null && p.anonymous) || (userId != null && p.user === userId)))
+        // eth addresses may be in checksumcase, but userId from server has no case
+        const userIdCaseInsensitive = typeof userId === 'string' ? userId.toLowerCase() : userId
+        return permissions.find((p) => (
+            p.operation === operation
+            && (
+                (userId == null && p.anonymous)
+                || (userId != null && p.user.toLowerCase() === userIdCaseInsensitive)
+            )
+        ))
     }
 
     grantPermission(operation, userId) {
         const permissionObject = {
             operation,
         }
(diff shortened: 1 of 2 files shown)


kare:
@timoxley Permission change is not yet merged and only thing with case-sensitivity that was recently merged is this https://github.com/streamr-dev/engine-and-editor/pull/740
Not certain if it is related.

src/rest/domain/Stream.js
```diff
@@ -47,16 +47,25 @@ export default class Stream {
 
     async hasPermission(operation, userId) {
         const permissions = await this.getPermissions()
-        return permissions.find((p) => p.operation === operation && ((userId == null && p.anonymous) || (userId != null && p.user === userId)))
+        // eth addresses may be in checksumcase, but userId from server has no case
+        const userIdCaseInsensitive = typeof userId === 'string' ? userId.toLowerCase() : userId
+        return permissions.find((p) => (
```
mirpo: why not smth like this:
```
res = undefined // or what is expected if not found?
if ((userId == null && p.anonymous) || (userId != null && p.user.toLowerCase() === userIdCaseInsensitive)) {
res = permissions.find((p) =>p.operation === operation) 
}
return res
```
now this construction is unreadable

timoxley: Agreed this could read better, though my change was simply reformatting what was already there to fit under the max line limit after adding `.toLowerCase` and `s/userId/userIdCaseInsensitive`

timoxley:
> only thing with case-sensitivity that was recently merged is this streamr-dev/engine-and-editor#740
> Not certain if it is related.

That change looks very related. 

@jtakalai what do you think, should we be storing the addresses in checksum case, or perhaps we should convert them back into checksum case when we read them?

src/rest/domain/Stream.js
timoxley: @mirpo changed in https://github.com/streamr-dev/streamr-client-javascript/pull/126/commits/faba92fba0b2e5e55b93f93751639e55616abfbd

src/rest/domain/Stream.js
```diff
@@ -46,17 +46,27 @@ export default class Stream {
     }
 
     async hasPermission(operation, userId) {
+        // eth addresses may be in checksumcase, but userId from server has no case
+        const userIdCaseInsensitive = typeof userId === 'string' ? userId.toLowerCase() : userId
```
jtakalai: what else could it be besides string and null? Maybe at least add comment in what cases you expect it go through as "just" userId

timoxley: forced to undefined if failing typeof string in 613ea4a

PR #129, opened 2020-03-26
Await before disconnect and update all libs

mirpo:


test/unit/StreamrClient.test.js
```diff
@@ -118,11 +118,14 @@ describe('StreamrClient', () => {
...
         c.send = (msgToSend) => {
-            const next = c.expectedMessagesToSend.shift()
-            if (!next) {
-                throw new Error(`Sending unexpected message: ${JSON.stringify(msgToSend)}`)
-            }
-            next.verificationFunction(msgToSend, next.msgToExpect)
+            return new Promise((resolve, reject) => {
+                const next = c.expectedMessagesToSend.shift()
+                if (!next) {
+                    throw new Error(`Sending unexpected message: ${JSON.stringify(msgToSend)}`)
```
timoxley, line 124: Note this doesn't cause the promise to reject, the promise hasn't even been created yet. This causes `c.send` to throw synchronously.

Might consider just changing it to an async function i.e. `c.send = async (msgToSend) => ` this will return a promise and any errors will always be converted to promise rejections

src/Connection.js
```diff
@@ -129,14 +129,26 @@ class Connection extends EventEmitter {
...
+                })
+
+                if (process.browser) {
+                    resolve()
+                }
+            } catch (err) {
+                this.emit('error', err)
+                reject(err)
+            }
+        })
```
timoxley, line 151: 👍

src/StreamrClient.js
```diff
@@ -778,7 +779,10 @@ export default class StreamrClient extends EventEmitter {
             const request = SubscribeRequest.create(sub.streamId, sub.streamPartition, sessionToken)
             debug('_requestSubscribe: subscribing client: %o', request)
             sp.setSubscribing(true)
-            this.connection.send(request)
+            await this.connection.send(request).catch((err) => {
+                sub.setState(Subscription.State.unsubscribed)
+                this.emit('error', `Failed to send subscribe request: ${err}`)
```
timoxley, line 784: Hm, should emit Error objects otherwise you lose stack trace.
```suggestion
                const error = new Error(`Failed to send subscribe request: ${err}`)
                error.stack = err.stack // this isn't ideal because it loses the above error's stack but not much you can do
                this.emit('error', err)
```

timoxley: Or do something like:

```suggestion
                debug(`Failed to send subscribe request: ${err}`)
                this.emit('error', err)
```
This might be better.

src/StreamrClient.js
```diff
@@ -789,9 +793,13 @@ export default class StreamrClient extends EventEmitter {
...
     }
 
-    _requestUnsubscribe(sub) {
+    async _requestUnsubscribe(sub) {
         debug('Client unsubscribing stream %o partition %o', sub.streamId, sub.streamPartition)
-        this.connection.send(UnsubscribeRequest.create(sub.streamId, sub.streamPartition))
+        const unsubscribeRequest = UnsubscribeRequest.create(sub.streamId, sub.streamPartition)
+        await this.connection.send(unsubscribeRequest).catch((err) => {
+            sub.setState(Subscription.State.subscribed)
+            this.handleError(`Failed to send unsubscribe request: ${err}`)
```
timoxley, line 801: similar thing here.

src/StreamrClient.js
```diff
@@ -819,7 +827,9 @@ export default class StreamrClient extends EventEmitter {
 
         if (request) {
             debug('_requestResend: %o', request)
-            this.connection.send(request)
+            await this.connection.send(request).catch((err) => {
+                this.handleError(`Failed to send resend request: ${err}`)
```
timoxley, line 831: and here

src/StreamrClient.js
```diff
@@ -819,7 +827,9 @@ export default class StreamrClient extends EventEmitter {
...
+            })
         } else {
             this.handleError("Can't _requestResend without resendOptions")
```
timoxley, line 834: oh no, it's how everything is

ok never mind, just do whatever

PR #131, opened 2020-03-26
Add streamr-client version in HTTP Headers

timoxley:
Adds header like so for server:
```
'User-Agent': 'Node v10.19.0 streamr-client-javascript/3.2.0dev'
```

`dev` only added if requests sent via non-npm-installed version of streamr-client.

Didn't append the `node-fetch` header that's currently sent because node-fetch doesn't expose it for extension, can only replace. This seems fine.

Client currently doesn't run in the browser (3.2.0), so haven't yet tested browser incantation, but intention is to append same `streamr-client-javascript/3.2.0dev` string to existing `navigator.userAgent`.

### DIFF
--- a/src/StreamrClient.js
+++ b/src/StreamrClient.js
@@ -32,7 +32,7 @@ import SubscribedStreamPartition from './SubscribedStreamPartition'
 import Stream from './rest/domain/Stream'
 import FailedToPublishError from './errors/FailedToPublishError'
 import MessageCreationUtil from './MessageCreationUtil'
-import { waitFor } from './utils'
+import { waitFor, getVersionString } from './utils'
 import RealTimeSubscription from './RealTimeSubscription'
 import CombinedSubscription from './CombinedSubscription'
 import Subscription from './Subscription'
@@ -92,6 +92,9 @@ export default class StreamrClient extends EventEmitter {
                 this.options.url = `${this.options.url}&messageLayerVersion=31`
             }
         }
+
+        // always add streamrClient version
+        this.options.url = `${this.options.url}&streamrClient=${getVersionString()}`
 
         // Backwards compatibility for option 'authKey' => 'apiKey'
         if (this.options.authKey && !this.options.apiKey) {
--- a/src/rest/DataUnionEndpoints.js
+++ b/src/rest/DataUnionEndpoints.js
@@ -21,7 +21,7 @@ import debugFactory from 'debug'
 
 import * as DataUnion from '../../contracts/DataUnion.json'
 
-import authFetch from './authFetch'
+import authFetch, { DEFAULT_HEADERS } from './authFetch'
 
 const { BigNumber, computeAddress, getAddress } = ethersUtils
 
@@ -51,9 +51,15 @@ function sleep(ms) {
     })
 }
 
-async function get(client, dataUnionContractAddress, endpoint, ...opts) {
+async function get(client, dataUnionContractAddress, endpoint, opts = {}) {
     const url = `${client.options.restUrl}/communities/${dataUnionContractAddress}${endpoint}`
-    const response = await fetch(url, ...opts)
+    const response = await fetch(url, {
+        ...opts,
+        headers: {
+            ...DEFAULT_HEADERS,
+            ...opts.headers,
+        },
+    })
     const json = await response.json()
     // server may return things like { code: "ConnectionPoolTimeoutException", message: "Timeout waiting for connection from pool" }
     //   they must still be handled as errors
--- a/src/rest/authFetch.js
+++ b/src/rest/authFetch.js
@@ -2,6 +2,17 @@ import fetch from 'node-fetch'
 import debugFactory from 'debug'
 
 import AuthFetchError from '../errors/AuthFetchError'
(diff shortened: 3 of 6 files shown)


timoxley:
After doing some manual testing in browser, turns out you cannot change the user-agent header, at least not in Chrome. It was originally a "forbidden" header, but when it was made not forbidden in 2015, it seems Chrome never updated to fix this.

Related bug in Chrome:
https://bugs.chromium.org/p/chromium/issues/detail?id=571722
Fixed bug in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1188932

@hpihkala let's pick a different http header and use that e.g. `Streamr-Version`  or perhaps the more restful approach would be to bake it into the client "Accepts" header e.g. `Accepts: "application/streamr-client+json;version=3.2.1, application/json"`

Also note there's no API for setting custom headers on the request for the WebSocket in the browser: https://stackoverflow.com/a/4361358/62851

hpihkala:
Ok, let’s do `X-Streamr-Client` for the http requests?

For ws, let’s pass the same version string as a query param (e.g. `streamr-client`) on the connect url. We’re already passing query params to ask for a specific protocol version.

timoxley:
@hpihkala we can add the `X-` however it's no longer necessary and the `Streamr-` prefix ensures no possible collisions with other headers:

> Custom proprietary headers have historically been used with an X- prefix, but this convention was deprecated in June 2012
 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

hpihkala:
Okay, without the `X-` then! I still see a lot of that, maybe people still use it because it’s easy to spot, especially if the headers are show in alphabetic order. I’m ok without the convention. Would still call it `Streamr-Client` or `Streamr-Library` instead of the more vague `Streamr-Version`.

timoxley:
heh, so looks like we might need to change nginx and/or engine-and-editor to whitelist the Streamr-Client header for CORS requests:

image

> Access to fetch at 'https://streamr.network/api/v1/streams/7wa7APtlTq6EC5iTCBy6dw' from origin 'http://localhost:8080' has been blocked by CORS policy: Request header field streamr-client is not allowed by Access-Control-Allow-Headers in preflight response.

I tried adding the correct header to the OPTIONS response in engine-and-editor, but it doesn't seem to be used, instead the OPTIONS response comes back from nginx:

```
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: origin, authorization, accept, content-type, x-requested-with
Access-Control-Allow-Methods: GET, HEAD, POST, PUT, DELETE, TRACE, OPTIONS
Access-Control-Allow-Origin: http://localhost:8080
Access-Control-Max-Age: 3600
Connection: keep-alive
Content-Length: 0
Date: Wed, 01 Apr 2020 15:38:05 GMT
Server: nginx/1.10.3
```

I don't see any cors-related config in streamr-docker-dev's nginx proxy so it's possible we'll need to add entirely new config in order to override the default behaviour.

timoxley:
Actually, that test was connecting to the production servers, so of course changes to nginx/engine-and-editor weren't being reflected. Sending a PR.

hpihkala:
@timoxley the `engine-and-editor` change is now deployed in all environments including production. Restarting CI tests on latest commit.

timoxley:
The CI failures on this are strange, gives a 404 from http://localhost:8081/streamr-core/api/v1/login/apikey

Only this branch consistently fails with this error, so I'm thinking there might actually be something up with e&e, not just a CI timing issue. Trying replicate locally.

```
FAIL test/integration/Session.test.js
  ● Console

    console.error src/StreamrClient.js:149
      AuthFetchError: Request to http://localhost:8081/streamr-core/api/v1/login/apikey returned with error code 404. <!doctype html><html lang="en"><head><title>HTTP Status 404 – Not Found</title><style type="text/css">H1 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:22px;} H2 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:16px;} H3 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:14px;} BODY {font-family:Tahoma,Arial,sans-serif;color:black;background-color:white;} B {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;} P {font-family:Tahoma,Arial,sans-serif;background:white;color:black;font-size:12px;}A {color : black;}A.name {color : black;}HR {color : #525D76;}</style></head><body><h1>HTTP Status 404 – Not Found</h1><hr class="line" /><p><b>Type</b> Status Report</p><p><b>Message</b> &quot;/index.gsp&quot; not found.</p><p><b>Description</b> The origin server did not find a current representation for the target resource or is not willing to disclose that one exists.</p><hr class="line" /><h3>Apache Tomcat/7.0.94</h3></body></html>
          at authFetch (/home/runner/work/streamr-client-javascript/streamr-client-javascript/src/rest/authFetch.j
```

timoxley:
Replicated locally, found the culprit, was a mistake in the PR (0f0174d). Fixed.

PR #132, opened 2020-03-27
Make EncryptionUtil Browser Compatible

timoxley:
Current version of `streamr-client` (3.2.0) does not work in the browser because `crypto-browserify`, which is used by webpack to shim nodes `crypto` module in the browser, doesn't implement the `generateKeyPairSync` function, or even the async `generateKeyPair`.

I have added a possibly browser-compatible implementation of `_generateKeyPair` using https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey. 

Until we set up a system for running in-browser tests, I am not sure this actually works yet.

The browser API does not provide a sync version, so I've made both versions async and added some handling to ensure it doesn't enqueue the same work twice. The async version should be used in node anyway since keygen is expensive and we don't want to block unnecessarily while it happens.

An alternative could be to use a JS implementation like https://www.npmjs.com/package/keypair, however:

> Performance greatly depends on the bit size of the generated private key. With 1024 bits you get a key in 0.5s-2s, with 2048 bits it takes 8s-20s, on the same machine. As this will block the event loop while generating the key, make sure that's ok or to spawn a child process or run it inside a webworker.

We're using a 4096 bit key and on my machine that takes 26 seconds to generate. So a pure JS version is not a viable option for the browser without involving the complexity of adding web workers, the calls become async with this option too. We should probably be using the native crypto apis.

The browser crypto APIs have another caveat though: they are only available in secure contexts (i.e. HTTPS). So the streamr-client won't be able to generate keys on any non-https page.  We should probably fail gracefully when this situation is encountered e.g. toggle on `window.isSecureContext`. 
Note it does *appear* to work with `localhost` or `127.0.0.1`, but will fail when deployed or accessed via a different hostname.

----

update: So rather than booting a browser and testing that way, I've set it up to use a webcrypto polyfill that works in node so the tests can run. This is a lot simpler and probably gives us most of the confidence of running it in a real browser. Also added a test step that confirms that the built client lib code can actually be used in another webpack build e.g. no unresolvable deps etc.

### DIFF
--- a/src/EncryptionUtil.js
+++ b/src/EncryptionUtil.js
@@ -2,7 +2,7 @@ import crypto from 'crypto'
 import util from 'util'
 
 // this is shimmed out for actual browser build allows us to run tests in node against browser API
-import WebCrypto from 'node-webcrypto-ossl'
+import { Crypto } from 'node-webcrypto-ossl'
 import { ethers } from 'ethers'
 import { MessageLayer } from 'streamr-client-protocol'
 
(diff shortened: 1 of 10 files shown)


mthambipillai:
The fix looks good to me but some unit tests are now failing. Plus it would be good to test that the generated keys are compatible. We should test that we can still use the same encrypt/decrypt methods with the keys generated in the browser. Also we should test it in the [e2e testing tool](https://github.com/streamr-dev/streamr-client-testing) against the Java client.

I'll look into if I have time before I leave but I'm not sure I'll manage..

src/EncryptionUtil.js
```diff
@@ -20,8 +34,18 @@ export default class EncryptionUtil {
...
+        return this._generateKeyPair()
+    }
+
+    isReady() {
+        return !!this.privateKey
+    }
+
     // Returns a Buffer
     decryptWithPrivateKey(ciphertext, isHexString = false) {
+        if (!this.isReady()) { throw new Error('EncryptionUtil not ready.') }
```
mthambipillai: This makes unit tests fail

timoxley: Fixed.

timoxley:
@mthambipillai 

> Plus it would be good to test that the generated keys are compatible. We should test that we can still use the same encrypt/decrypt methods with the keys generated in the browser. 

Agreed. What do you think is the rough outline for how to test this?

> Also we should test it in the e2e testing tool against the Java client.

I believe this will require booting a browser.

PR #134, opened 2020-04-10
added wait time to accomodate 1m parity block time

isentropy:
parity needs some extra wait time before testing withdraw


timoxley:
> Does the block time HAVE to be so long though?

yeah given how slow everything already is I'd be reluctant to introduce any additional delays unless absolutely necessary.

timoxley:
I'm going to merge this so CI will come green

PR #135, opened 2020-04-16
DU-116: swap in the new DU contract

jtakalai:
no renaming community -> dataunion done here yet, just contract replacement. See https://streamr.atlassian.net/browse/DU-112 for names rectification


src/rest/CommunityEndpoints.js
```diff
@@ -19,7 +19,7 @@ import {
 } from 'ethers'
 import debugFactory from 'debug'
 
-import * as CommunityProduct from '../../contracts/CommunityProduct.json'
+import * as CommunityProduct from '../../contracts/DataunionVault.json'
```
timoxley, line 22: `Dataunion` vs `DataUnion`?

timoxley:
Also note new conflict.

src/rest/CommunityEndpoints.js
hpihkala: Here's what I've been using:
- `DataUnion` for class names
- `dataUnion` for variables
- In URL routes, just `dataunion` to avoid the annoyances of mixed case and punctuation

timoxley:
@jtakalai maybe we could avoid the need to have the contracts in the client altogether?

Perhaps all the places it talks to the contract object should actually be API calls to the DU server.

Currently the contracts require every user of the client to download them and though the client is already bloated we could avoid an additional 45kb by pushing all the contract talk into the DU server?

jtakalai:
StreamrClient should be able to talk directly to Ethereum. The contract size issue could maybe be solved by just taking the part of ABI that's needed, and/or downloading bytecode from elsewhere.

timoxley:
Closing as superceded by #143

PR #136, opened 2020-04-17
Rename nodejs dist file

hpihkala:
Some users tried to import `dist/streamr-client.min.js` in the browser and got confused.

As a quick fix I propose renaming the nodejs build to `dist/streamr-client.nodejs.js` to clearly indicate the target platform instead of omitting it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test": "test"
   },
   "scripts": {
-    "build": "NODE_ENV=production webpack --mode=production --progress",
+    "build": "rm dist/* && NODE_ENV=production webpack --mode=production --progress",
     "prepack": "npm run build",
     "dev": "webpack --progress --colors --watch --mode=development",
     "eslint": "eslint .",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "type": "git",
     "url": "git://github.com/streamr-dev/streamr-client.git"
   },
-  "main": "dist/streamr-client.js",
+  "main": "dist/streamr-client.nodejs.js",
   "browser": "dist/streamr-client.web.min.js",
   "directories": {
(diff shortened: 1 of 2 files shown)


mirpo:
Maybe we need to change a bit `package.json`.
So `main` will point to `browser` version and only `module` to `node.js` version.
Like here:
https://github.com/node-fetch/node-fetch/blob/6e1b579ce0385ac72f0c63f5011ffc10fea45c85/package.json#L5-L11

hpihkala:
> Maybe we need to change a bit `package.json`.
> So `main` will point to `browser` version and only `module` to `node.js` version.
> Like here:
> https://github.com/node-fetch/node-fetch/blob/6e1b579ce0385ac72f0c63f5011ffc10fea45c85/package.json#L5-L11

I actually can't find any documentation for the `module` entry - could you point us to it? 

Anyway, the current approach works, so why fix something that isn't broken?

mirpo:
I know only this good article https://2ality.com/2017/04/setting-up-multi-platform-packages.html

package.json
```diff
@@ -6,13 +6,13 @@
...
   },
-  "main": "dist/streamr-client.js",
+  "main": "dist/streamr-client.nodejs.js",
   "browser": "dist/streamr-client.web.min.js",
   "directories": {
     "test": "test"
   },
   "scripts": {
-    "build": "NODE_ENV=production webpack --mode=production --progress",
+    "build": "rm dist/* && NODE_ENV=production webpack --mode=production --progress",
```
timoxley, line 15: This breaks CI since on the first run `dist` doesn't exist.

hpihkala:
Ah crap, my bad. Will fix

PR #139, opened 2020-04-23
Really disconnect from server

mirpo:


src/StreamrClient.js
```diff
@@ -716,7 +716,11 @@ export default class StreamrClient extends EventEmitter {
      */
 
     async ensureDisconnected() {
-        if (this.isDisconnected()) { return Promise.resolve() }
+        this.connection.clearReconnectTimeout()
```
timoxley, line 719: I wonder if this should go into `disconnect()` instead

PR #144, opened 2020-05-26
Add failing test for pubsub across multiple clients.

timoxley:
WIP trying to nail down cause of CI failure.


timoxley:
Good news, this passes now.

mirpo:
Now it works

PR #145, opened 2020-05-29
Fixing CI

timoxley:
Increasing resend test timeout to maybe have it pass in CI on Node 10. Very curious why it only affects the Node 10 build, perhaps a performance difference?


timoxley:
🤷

PR #146, opened 2020-06-02
Add client testing tool

mikhaelsantos:


.github/workflows/nodejs.yml
```diff
@@ -48,4 +47,33 @@ jobs:
...
+        git clone https://github.com/streamr-dev/streamr-client-testing.git
+        cd ${GITHUB_WORKSPACE}/streamr-client-testing
+        echo "Setup packagejson to use PR JS client"
+        sed -i "s/com.streamr:client:1.3.0/com.streamr:client:+/g" build.gradle
+        sed -i "s/\"streamr-client\": \"\^3.1.2\"/\"streamr-client\":\"PATH\"/g" package.json
+        ## fix for escaping / of workspace path
+        sed -i "s|"PATH"|"${GITHUB_WORKSPACE}"|g" package.json
+        echo "Prepare for test"
+        ## npm install is used because package-lock.json could be form a previous client version.
+        npm install
```
mirpo, line 76: `npm install`? not `npm ci` ?

.github/workflows/nodejs.yml
```diff
@@ -48,4 +47,33 @@ jobs:
...
+        sudo service mysql stop
+        if [ ! -d streamr-docker-dev ]; then # Skip clone on subsequent attemps.
+         git clone https://github.com/streamr-dev/streamr-docker-dev.git
+        fi
+        sudo ifconfig docker0 10.200.10.1/24
+        ${GITHUB_WORKSPACE}/streamr-docker-dev/streamr-docker-dev/bin.sh start smart-contracts-init nginx engine-and-editor --wait
+        echo "Testing Tool setup"
+        git clone https://github.com/streamr-dev/streamr-client-testing.git
+        cd ${GITHUB_WORKSPACE}/streamr-client-testing
+        echo "Setup packagejson to use PR JS client"
```
mirpo, line 69: typo `package.json`

.github/workflows/nodejs.yml
```diff
@@ -48,4 +47,33 @@ jobs:
...
+        sed -i "s/com.streamr:client:1.3.0/com.streamr:client:+/g" build.gradle
+        sed -i "s/\"streamr-client\": \"\^3.1.2\"/\"streamr-client\":\"PATH\"/g" package.json
```
mirpo, line 71: it should work only with version `^3.1.2`?

timoxley:
@mikhaelsantos can you elaborate on what this does and why

.github/workflows/nodejs.yml
timoxley: > npm install is used because package-lock.json could be form a previous client version.

we should never be landing package.json dependency changes in any branch, in any repo, that don't also come with corresponding package-lock changes.

mikhaelsantos: @timoxley This is changing on the fly. 
Here the idea is to build streamr-client-testing-tool with the PR code, Is there any other way to do that using package-lock?

.github/workflows/nodejs.yml
mikhaelsantos: The testing client made by @mthambipillai has that value there. I don't know if it's essential or not. What is actually being gone there (if you check a bit lower) is changing the version to point to the local code.

mikhaelsantos:
> @mikhaelsantos can you elaborate on what this does and why

Run JS Client PR library against Dev environment.

The idea is to take @mthambipillai streamr-client-testing-tool, switch the JS streamr client with the PR code and test that is works against the Dev enviroment.

What are we testing is:
* That is that it works with the Java Streamr Client.
* That it is backward compatible with the clients in the dev environment.

PR #148, opened 2020-06-24
CORE-1945: Set Content-Type for requests

kare:
Including POST request with 0 body length.

### DIFF
--- a/src/rest/authFetch.js
+++ b/src/rest/authFetch.js
@@ -18,8 +18,9 @@ const authFetch = async (url, session, opts = {}, requireNewToken = false) => {
             ...opts.headers,
         }
     }
-    // add default 'Content-Type: application/json' header for requests with a body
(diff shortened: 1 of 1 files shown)


hpihkala:
Note failing tests!

timoxley:
the failing test is likely the original bug I noticed i.e. it's ignoring the passed in id and returning all items, I'm guessing a PR needs to be merged into the engine before this will pass here? e.g. https://github.com/streamr-dev/engine-and-editor/pull/795 ?

hpihkala:
So @kare could you confirm if this is the case: the tests of this PR depend on https://github.com/streamr-dev/engine-and-editor/pull/795, and vice versa?

kare:
@hpihkala @timoxley Yes, Engine requires this for REST API tests to pass and vice versa.

PR #150, opened 2020-07-13
Error Handling Cleanup

timoxley:
Error output is suppressed when expected. All errors should be caught, promises handled.

* No unhandledRejection errors, no "expected" error log messages.
* Prevents cases where the same error was logged twice.
* Log messages are prefixed/scoped to a client instance where possible.
* Added additional logging around http methods.
* Prevented proceeding blindly with inbox stream subscription if disconnected after connection.
* Stops background fetching of user info on `new Client`, only starts fetching if/when user first needs session info, which allows the error to be caught by this caller.

Doesn't fully fix everything, there are still errors that can fire in event handlers or setTimeout which have no way of being caught and handled properly yet.

### DIFF
--- a/src/AbstractSubscription.js
+++ b/src/AbstractSubscription.js
@@ -117,9 +117,9 @@ export default class AbstractSubscription extends Subscription {
     _requestGroupKeyAndQueueMessage(msg, start, end) {
         this.emit('groupKeyMissing', msg.getPublisherId(), start, end)
         const publisherId = msg.getPublisherId().toLowerCase()
-        this.nbGroupKeyRequests[publisherId] = this.nbGroupKeyRequests[publisherId] + 1 || 1
+        this.nbGroupKeyRequests[publisherId] = 1
         clearInterval(this.waitingForGroupKey[publisherId])
-        const timer = setInterval(() => {
+        this.waitingForGroupKey[publisherId] = setInterval(() => {
             if (this.nbGroupKeyRequests[publisherId] < MAX_NB_GROUP_KEY_REQUESTS) {
                 this.nbGroupKeyRequests[publisherId] += 1
                 this.emit('groupKeyMissing', msg.getPublisherId(), start, end)
@@ -128,7 +128,6 @@ export default class AbstractSubscription extends Subscription {
                 this._cancelGroupKeyRequest(publisherId)
             }
         }, this.propagationTimeout)
-        this.waitingForGroupKey[publisherId] = timer
         this._addMsgToQueue(msg)
     }
 
--- a/src/AbstractSubscription.js
+++ b/src/AbstractSubscription.js
@@ -117,7 +117,7 @@ export default class AbstractSubscription extends Subscription {
     _requestGroupKeyAndQueueMessage(msg, start, end) {
         this.emit('groupKeyMissing', msg.getPublisherId(), start, end)
         const publisherId = msg.getPublisherId().toLowerCase()
-        this.nbGroupKeyRequests[publisherId] = 1
+        this.nbGroupKeyRequests[publisherId] = 1 // reset retry counter
         clearInterval(this.waitingForGroupKey[publisherId])
         this.waitingForGroupKey[publisherId] = setInterval(() => {
             if (this.nbGroupKeyRequests[publisherId] < MAX_NB_GROUP_KEY_REQUESTS) {
--- a/src/AbstractSubscription.js
+++ b/src/AbstractSubscription.js
@@ -1,29 +1,26 @@
 import debugFactory from 'debug'
 import { Errors, Utils } from 'streamr-client-protocol'
+import uniqueId from 'lodash.uniqueid'
 
 import VerificationFailedError from './errors/VerificationFailedError'
 import Subscription from './Subscription'
 import UnableToDecryptError from './errors/UnableToDecryptError'
 
 const { OrderingUtil } = Utils
-const debug = debugFactory('StreamrClient::AbstractSubscription')
-
-const defaultUnableToDecrypt = (error) => {
-    const ciphertext = error.streamMessage.getSerializedContent()
-    const toDisplay = ciphertext.length > 100 ? `${ciphertext.slice(0, 100)}...` : ciphertext
-    console.warn(`Unable to decrypt: ${toDisplay}`)
-}
 
 const MAX_NB_GROUP_KEY_REQUESTS = 10
 
 export default class AbstractSubscription extends Subscription {
-    constructor(streamId, streamPartition, callback, groupKeys, propagationTimeout, resendTimeout, orderMessages = true,
-        onUnableToDecrypt = defaultUnableToDecrypt) {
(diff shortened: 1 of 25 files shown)


src/Subscription.js
```diff
@@ -2,28 +2,34 @@ import EventEmitter from 'eventemitter3'
...
             throw new Error('No callback given!')
         }
         this.streamId = streamId
         this.streamPartition = streamPartition
         this.callback = callback
-        this.id = uniqueId('sub')
+        const id = uniqueId('sub')
+        this.id = id
+        if (debug) {
+            this.debug = debug.extend(this.constructor.name).extend(id)
```
hpihkala, line 25: Will `this.constructor.name` get changed to something intelligible by minification? I've encountered that sometimes when trying to use `this.constructor.name`, not sure if it's a problem with current tooling.

test/unit/MessageCreationUtil.test.js
```diff
@@ -506,7 +506,7 @@ describe('MessageCreationUtil', () => {
                 requestId,
             })
 
-            expect(streamMessage.getStreamId()).toBe('destinationAddress'.toLowerCase()) // sending to subscriber's inbox stream
+            expect(streamMessage.getStreamId()).toBe('destinationAddress') // sending to subscriber's inbox stream
```
hpihkala, line 509: Should the `streamId` actually be lowercase?

timoxley: `system/keyexchange` vs `SYSTEM/keyexchange`

src/Connection.js
```diff
@@ -22,7 +27,7 @@ class Connection extends EventEmitter {
         this.emit(state)
     }
 
-    connect() {
+    async connect() {
         if (this.state === Connection.State.CONNECTING) {
             return Promise.reject(new Error('Already connecting!'))
```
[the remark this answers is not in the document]
timoxley: correct

src/AbstractSubscription.js
```diff
@@ -97,20 +110,21 @@ export default class AbstractSubscription extends Subscription {
...
-            console.warn('Failed to decrypt. Requested the correct decryption key(s) and going to try again.')
+            this.debug('Failed to decrypt. Requested the correct decryption key(s) and going to try again.')
         }
     }
 
     _requestGroupKeyAndQueueMessage(msg, start, end) {
         this.emit('groupKeyMissing', msg.getPublisherId(), start, end)
         const publisherId = msg.getPublisherId().toLowerCase()
-        this.nbGroupKeyRequests[publisherId] = 1
+        this.nbGroupKeyRequests[publisherId] = this.nbGroupKeyRequests[publisherId] + 1 || 1
```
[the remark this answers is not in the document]
timoxley: This is a good point. I'll revert and add a comment.

src/Subscription.js
timoxley: Yeah it should be left as-is but I'll double-check

src/AbstractSubscription.js
```diff
@@ -97,20 +110,21 @@ export default class AbstractSubscription extends Subscription {
...
+        clearInterval(this.waitingForGroupKey[publisherId])
         const timer = setInterval(() => {
             if (this.nbGroupKeyRequests[publisherId] < MAX_NB_GROUP_KEY_REQUESTS) {
                 this.nbGroupKeyRequests[publisherId] += 1
                 this.emit('groupKeyMissing', msg.getPublisherId(), start, end)
             } else {
-                console.warn(`Failed to receive group key response from ${publisherId} after ${MAX_NB_GROUP_KEY_REQUESTS} requests.`)
+                this.debug(`WARN: Failed to receive group key response from ${publisherId} after ${MAX_NB_GROUP_KEY_REQUESTS} requests.`)
```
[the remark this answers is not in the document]
timoxley: `onUnableToDecrypt` doesn't do much useful, just logs the error. Agreed it should all run through the same failure code path but let's defer that as this stuff will likely change with the upcoming groupkey handling revision.

src/Connection.js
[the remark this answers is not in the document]
timoxley: fixed

src/StreamrClient.js
```diff
@@ -227,35 +234,41 @@ export default class StreamrClient extends EventEmitter {
...
             } else {
-                debug('resent: Subscription %s is gone already', response.requestId)
+                this.debug('resent: Subscription %s is gone already', response.requestId)
             }
         })
 
         // On connect/reconnect, send pending subscription requests
         this.connection.on('connected', async () => {
-            debug('Connected!')
+            await new Promise((resolve) => setTimeout(resolve, 0)) // wait a tick
```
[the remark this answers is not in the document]
timoxley: fixed

src/StreamrClient.js
```diff
@@ -290,40 +303,50 @@ export default class StreamrClient extends EventEmitter {
...
 
+    /**
+     * Override to control output
+     */
+
+    onError(error) { // eslint-disable-line class-methods-use-this
+        console.error(error)
+    }
+
     async _subscribeToInboxStream() {
```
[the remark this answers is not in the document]
timoxley: Done

src/Subscription.js
timoxley: Seems to keep the class name intact.

src/AbstractSubscription.js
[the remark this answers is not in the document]
timoxley: Fixed.

test/unit/MessageCreationUtil.test.js
hpihkala: Ah yes, of course. I think this expectation here was a remnant of the key exchange stream id being just the address. The mixed-case problem hasn't gone anywhere though, so both `0xabc` and `0xABC` should both result in the key exchange messages getting sent to `SYSTEM/keyexchange/0xabc`, never `SYSTEM/keyexchange/0xABC`. I wonder if there's test coverage for this elsewhere?

src/MessageCreationUtil.js
```diff
@@ -214,7 +214,7 @@ export default class MessageCreationUtil {
             streamId,
             requestId,
         }
-        const [messageId, prevMsgRef] = this.createDefaultMsgIdAndPrevRef(destinationAddress.toLowerCase(), publisherId)
+        const [messageId, prevMsgRef] = this.createDefaultMsgIdAndPrevRef(destinationAddress, publisherId)
```
hpihkala, line 217: Noticed one more cosmetic thing here: the variable name `destinationAddress` is a remnant from the time when the key exchange streamId was just the address. Now that it's `SYSTEM/keyexchange/address`, the variable should be renamed to `keyExchangeStreamId` or similar.

Similar misleading naming is currently happening in other message creation methods here too.

timoxley: Changed `destinationAddress` to `keyExchangeStreamId` in 0b57615.

PR #156, opened 2020-08-12
1. Remove Encryption, Refactor Client

timoxley:
Removed encryption logic in preparation for re-adding the encryption/decryption logic in a more "pipeline"-like architecture.

Did some substantial refactoring of the client to try isolate related pieces of logic & state: Connection, Publishing, Subscribing + Resends. Flattened some logic and moved event handlers into methods so it's possible to actually remove them, but otherwise the code is largely unchanged, things are just moved around a bit.

Followed by:
* Connection/Reconnection Refactoring https://github.com/streamr-dev/streamr-client-javascript/pull/160
* Publish/Subscribe Promisification https://github.com/streamr-dev/streamr-client-javascript/pull/162
* Publish Refactoring https://github.com/streamr-dev/streamr-client-javascript/pull/164


mirpo (approved):
LGTM

src/Subscriber.js
```diff
@@ -0,0 +1,382 @@
...
+            stream.getSubscriptions().forEach((sub) => {
+                this._removeSubscription(sub)
+                sub.setState(Subscription.State.unsubscribed)
+            })
+        }
+
+        this._checkAutoDisconnect()
+    }
+
+    async onClientConnected() {
```
harbu, line 70: `async` needed?

src/Subscriber.js
```diff
@@ -0,0 +1,382 @@
...
+                this.subscribedStreamPartitions[key].getSubscriptions().forEach((sub) => {
+                    if (sub.getState() !== Subscription.State.subscribed) {
+                        this._resendAndSubscribe(sub).catch((err) => {
+                            this.client.emit('error', err)
+                        })
+                    }
+                })
+            })
+        } catch (err) {
+            this.client.emit('error', err)
```
harbu, line 84: Since `_resendAndSubscribe` has a catch above, seems like anything that would fall here would be a "genuine" error worth throwing?

harbu (approved):
LGTM; Skimmed through since it was a pretty big PR.

A few very minor comments. Prolly were in the code before these changes so feel free to ignore.

One thing that I took note of with the current code base: lots of state handling, registering / unregistering, bookkeeping, connection management etc. Wonder how much of this is really necessary and how much a result of accidental complexity.

src/Resender.js
```diff
@@ -0,0 +1,188 @@
...
+        const sub = this.resendUtil.getSubFromResendResponse(response)
+
+        if (!stream || !sub || !stream.getSubscription(sub.id)) {
+            this.debug('resent: Subscription %s is gone already', response.requestId)
+            return
+        }
+        stream.getSubscription(sub.id).handleResending(response)
+    }
+
+    async onUnicastMessage(msg) {
```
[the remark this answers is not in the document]
timoxley: right, doesn't help for event handlers

src/Subscriber.js
timoxley: yep correct, good spotting

src/Publisher.js
```diff
@@ -0,0 +1,131 @@
...
+        client.on('connected', this.onConnected)
+    }
+
+    async onConnected() {
+        if (!this.client.isConnected()) { return }
+        try {
+            // Check pending publish requests
+            const publishQueueCopy = this.publishQueue.slice(0)
+            this.publishQueue = []
+            publishQueueCopy.forEach((publishFn) => publishFn())
```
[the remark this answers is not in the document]
timoxley: Yeah that's probably better, though this is the old-code as-is.

I have a 'queue' in the new version but it isn't clamped at a max value like this, just works on promises. Should probably get a proper queue again in order to limit the max outstanding.

timoxley:
This all lives in the `5.x` branch and PR #192 now.

PR #157, opened 2020-08-17
Remove createSecret argument 'secret'

jtakalai:
I vaguely remember setting your own secret was removed from backend... or at least it shouldn't be possible through the "normal flow" with streamr-client?

I'm opening this PR because these changes have been made long time ago and forgotten, but they look reasonable. I'd like a second opinion on them.


timoxley (approved):
lgtm.

PR #167, opened 2020-10-02
Streamr Client 5.0.0

timoxley:
This is basically a new streamr client with rebuilt internals, including a message subscription “pipeline”, more robust/understandable/consistent connection handling and a lot more tests + reliability around connection & timing issues.

External interface is largely the same, but timings may differ and there may be some edge cases not covered by tests.

This is built on top of, and supersedes changes in #156, #160, #162, #164 & #166

#### Rough WIP Changelog

* `subscribe`/`resend` methods are now async.
* No more `subscribed` event (`subscribe()` promise now resolves when this would have fired)
* No more `no_resend` event, only `resent` whether there were messages or not.
* Subscribe, resend & subscribe with resend all use the same machinery, so their behaviour should be identical.
* Added `collect()` function to subscriptions that collects the results into an array of messages and resolves when the subscription ends, mainly/only useful for resends.
* Subscriptions are async iterators. You can access this interface if you don’t supply an `onMessage` handler to `subscribe`/`resend`.
* Connection emits `done` when it's been disconnected and will not reconnect.
* Connection emits `connected` only when it's connected and will not be immediately disconnecting. 
* autoConnect + autoDisconnect options become forcibly disabled if you explicitly call `connect()` or `disconnect()`. This is because I could not come up a sensible, consistent expected behaviour that combined explicit & automatic connect/disconnect. Can reenable auto behaviour without creating a new client or mutating options using `enableAutoConnect()`/`enableAutoDisconnect()`. This means if you `connect` you must `disconnect` or `enableAutoDisconnect()` to close the connection.
* `connect()` & `disconnect()` no longer error if already {dis}connecting/{dis}connected. This is basically what `ensureConnected`/`ensureDisconnected` provided before.
* The `ensureConnected`/`ensureDisconnected`  methods still exist but are now just aliases of `connect` & `disconnect`.
* Uses latest alpha version of streamr protocol `7.0.0-alpha.2` to get these fixes: https://github.com/streamr-dev/streamr-client-protocol-js/pull/46 & https://github.com/streamr-dev/streamr-client-protocol-js/pull/47
* Random errors no longer disconnect the client.
* Reconnecting on unexpected disconnection should work reliably now, there's a hard distinction between user-initiated disconnects and those that happen by accident.
* Resends should work more reliably now.
* Cleaned up and improved logging. Is more verbose but should be clearer what’s going on.
* Removed regenerator babel transform.
* Almost all functionality is now tested via integration tests, rather than implementation-coupled+heavily mocked unit tests.
* Updated all tests to not use `async (done) =>` in preparation for `jest-circus` jest runner being enabled by default in Jest 27+. Enabled `jest-circus` by default anyway. This fixes an issue where if a test fails asynchronously, the `afterEach`/`afterAll` hooks would run unchained i.e. `beforeEach` for next test would start before `afterEach` was complete. This led to some very confusing test results at times.
* Moved data union tests into “flakey” test folder. These tests still run, but won’t fail CI.


timoxley:
Going to make this into the new `5.x` branch. 👍

PR #168, opened 2020-10-15
Only add Content-Type header for POST and PUT

hpihkala:
Previous code lead to query params being ignored on `GET` requests. 

While the backend would ideally be more tolerant to "abuse", it's harder to fix on that end because Grails conforms to certain conventions which are hard to customize. 

The RFC is unclear about whether it's wrong to add `Content-Type` to requests that can't even have a body, but in any case I don't think we should add headers that have _no chance of doing anything except perhaps throwing off the backend_.

One thing that would fail for example was:

```
client.listStreams({
    name: 'foo'
})
```

The above is expected to return an empty list, if the user doesn't have a stream named `foo`. However, it currently returns ALL streams which the user can access, because the filter criteria is passed as query params (as per API spec) and gets ignored.

There was an integration test that would have caught this, but it had been skipped. :( This PR fixes the issue and adds the test back in.

### DIFF
--- a/src/rest/authFetch.js
+++ b/src/rest/authFetch.js
@@ -18,9 +18,8 @@ const authFetch = async (url, session, opts = {}, requireNewToken = false) => {
             ...opts.headers,
         }
     }
-    // add default 'Content-Type: application/json' header for all requests
-    // including 0 body length POST calls
-    if (!options.headers['Content-Type']) {
+    // add default 'Content-Type: application/json' header for all POST and PUT requests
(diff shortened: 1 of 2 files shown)


timoxley (approved):
lgtm.

hpihkala:
Lol, there actually had been an integration test that would have caught the issue, but it was skipped since [this commit](https://github.com/streamr-dev/streamr-client-javascript/commit/f7f645d9878a642146aeb90c595b294ce9478924).

Added that test back in.

PR #175, opened 2020-11-20
DU-33 transport signatures upon Data Union withdraw

jtakalai:


jtakalai:
There are tests not passing now in CI, presumably because it does merge with DU-2 first, and that branch is broken right now, waiting for getStats to return 6 elements instead of 5 (and this change being updated to dockers)

jtakalai:
Now it's same as base DU-2 branch, signed withdraw failing

PR #176, opened 2020-12-03
Add Ethereum address validation to DU functions

jtakalai:


src/rest/DataUnionEndpoints.js
```diff
@@ -526,14 +526,17 @@ async function getSidechainContractReadOnly(client, options = {}) {
 // //////////////////////////////////////////////////////////////////
 
 export async function calculateDataUnionMainnetAddress(dataUnionName, deployerAddress, options) {
-    return getDataUnionMainnetAddress(this, dataUnionName, deployerAddress, options)
+    const a = getAddress(deployerAddress) // throws if bad address
```
juhah, line 529: I would prefer named variables, eg. `const address = getAddress(deployerAddress)`

juhah (approved):
Good to have the admin fee fractions converted in the client 👍

PR #178, opened 2020-12-07
Bug: getOrCreateStream for non-existing stream

teogeb:
```getOrCreateStream``` doesn't create a stream if it is called with an ```id``` parameter.

```getOrCreateStream``` can be called with an ```id``` parameter or with a name parameter, and as the name suggests it either gets and existing stream or creates a new one. All three other use cases work ok:
- get stream by name
- create a new stream by name
- get stream by id

The problem is in ```getStream``` which throws an error as it gets HTTP 404 error from the backend. There is no documentation what ```getStream``` should return if the stream doesn't exists (```undefined``` or throw an error). Most likely it should return an ```undefined```, because:

- ```getStreamByName``` also returns undefined in the same situation
- ```getOrCreateStream``` tries to checks whether getStream returns an ```undefined```: https://github.com/streamr-dev/streamr-client-javascript/blob/693f3f3945ae355e5edd2fbf93cb42591653f1a7/src/rest/StreamEndpoints.js#L96
- there is a ternary operator in ```getStream``` which tries to return ```undefined:``` https://github.com/streamr-dev/streamr-client-javascript/blob/693f3f3945ae355e5edd2fbf93cb42591653f1a7/src/rest/StreamEndpoints.js#L44 (it doesn't do that, because ```authFetch``` throws the error)

This PR modifies the ```getStream``` method to return ```undefined``` for non-existing streams. ```getOrCreateStream``` doesn't need any modifications as it can already handle the ```undefined``` result from getStream().

### DIFF
--- a/src/rest/StreamEndpoints.js
+++ b/src/rest/StreamEndpoints.js
@@ -46,9 +46,8 @@ export async function getStream(streamId) {
     } catch (e) {
         if (e.response && e.response.status === 404) {
             return undefined
-        } else { // eslint-disable-line no-else-return
-            throw e
         }
+        throw e
     }
 }
 
--- a/src/rest/StreamEndpoints.js
+++ b/src/rest/StreamEndpoints.js
@@ -40,8 +40,16 @@ export async function getStream(streamId) {
         streamId,
     })
     const url = getEndpointUrl(this.options.restUrl, 'streams', streamId)
-    const json = await authFetch(url, this.session)
-    return json ? new Stream(this, json) : undefined
+    try {
+        const json = await authFetch(url, this.session)
+        return new Stream(this, json)
+    } catch (e) {
+        if (e.response && e.response.status === 404) {
+            return undefined
+        } else { // eslint-disable-line no-else-return
+            throw e
+        }
+    }
 }
 
 export async function listStreams(query = {}) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/rest/StreamEndpoints.js [hunk 1]
```diff
@@ -40,8 +40,16 @@ export async function getStream(streamId) {
...
     const url = getEndpointUrl(this.options.restUrl, 'streams', streamId)
-    const json = await authFetch(url, this.session)
-    return json ? new Stream(this, json) : undefined
+    try {
+        const json = await authFetch(url, this.session)
+        return new Stream(this, json)
+    } catch (e) {
+        if (e.response && e.response.status === 404) {
+            return undefined
+        } else { // eslint-disable-line no-else-return
```
timoxley, line 49: Yeah just remove the else here

test/integration/StreamEndpoints.test.js
```diff
@@ -47,6 +47,20 @@ describe('StreamEndpoints', () => {
         })
     })
 
+    describe('get', () => {
```
timoxley, line 50:
```suggestion
    describe('getStream', () => {
```

timoxley (approved):
lgtm

PR #183, opened 2020-12-15
Add payForSignature option

jtakalai:
set to true by default, so that it will work in both scenarios: we must pay or bridge pays (and we try to also pay but fail)


src/rest/DataUnionEndpoints.js
```diff
@@ -448,7 +450,9 @@ async function untilWithdrawIsComplete(client, getWithdrawTxFunc, getBalanceFunc
         log(`Waiting until sidechain AMB has collected required signatures for hash=${messageHash}...`)
         await until(async () => requiredSignaturesHaveBeenCollected(client, messageHash, options), pollingIntervalMs, retryTimeoutMs)
         log(`Transporting signatures for hash=${messageHash}`)
```
isentropy, line 452: shouldnt this log line be in if() block below?

src/rest/DataUnionEndpoints.js
```diff
@@ -448,7 +450,9 @@ async function untilWithdrawIsComplete(client, getWithdrawTxFunc, getBalanceFunc
         log(`Waiting until sidechain AMB has collected required signatures for hash=${messageHash}...`)
```
isentropy, line 450: if the message is alreadySent, don't throw an error. Just print log line. This happens on xDai-mainnet bridge for ex.

PR #184, opened 2020-12-16
escape forward slashes in streamId when calling publishHttp

harbu:
Escape forward slahes in streamId url parameter when using `publishHttp` to publish messages


teogeb:
If you use `getEndpointUrl(this.options.restUrl, 'streams', streamId, 'data')`, the method encodes the path segments automatically. We could also fix `createStream` to use getEndpointUrl(this.options.restUrl, 'streams')` for consistency.

PR #165, opened 2020-09-16
StreamrClient functions for new Data Unions

jtakalai:

PR #188, opened 2021-01-06
Client 5.x + DataUnions v2

timoxley:
Merged  DU2 branch #165 into 5.x + encryption branch #185 + added some merge fixes.


src/Session.js
```diff
@@ -1,18 +1,23 @@
 import EventEmitter from 'eventemitter3'
-import { ethers } from 'ethers'
+import { Wallet } from '@ethersproject/wallet'
+import { Web3Provider } from '@ethersproject/providers'
 
 export default class Session extends EventEmitter {
-    constructor(client, options) {
+    constructor(client, options = {}) {
```
jtakalai, line 6: Sorry about these changes  :D  but they were kind of necessary to use the "new" window.ethereum, or maybe it was just the signing refactor. I would've wanted to do more, actually, but maybe I will get to it later.

jtakalai: and by signing I mean signing transactions generated by the Data Union endpoints. I *know* I should have left all other parts alone, such as message signing (maybe I didn't commit those changes after all?) but it was done using the same credentials and that code that directly accessed the provider was there copy-pasted. In the end it seems I didn't replace that code but the "cleanups" went in there nevertheless. Hope it wasn't too much trouble.

src/user/index.js
```diff
@@ -0,0 +1,50 @@
+import { computeAddress } from '@ethersproject/transactions'
```
jtakalai, line 1: nice, these things were copy-pasted here and there (maybe still are? I think Session.js and Signer.js), good to get them in one place.

jtakalai (commented):
That matrix of tests looks pretty banger in the checks tab :D  I guess the goal is still to have them run in decent time? DU tests aren't very great at that I'm afraid  /:

jtakalai (approved):

src/Session.js
timoxley: Not a major issue. Just took a bit to figure out what changed and what didn't.

src/StreamrClient.js
```diff
@@ -372,11 +310,29 @@ export default class StreamrClient extends EventEmitter {
         return this.connection.enableAutoDisconnect(...args)
     }
 
+    getAddress() {
```
jtakalai, line 313: maybe instead of these functions, everyone who wants those things could just call `client.ethereum.getSigner()` etc.? Not all of that is really relevant to most user needs, so no need to wrap more than just `getAddress()` maybe

jtakalai (approved):
I think this is a good refactor.

Note also that there's signing happening in Signer, and they use Ethereum credentials / signing methods. I'm not entirely sure why we even have "custom" SigningUtils...

src/StreamrClient.js
timoxley: Good thought! Let's do that.

PR #197, opened 2021-02-09
NET-178: Storage node methods

teogeb:
Methods to query+modify storage nodes. Two new entity classes: `StreamPart` (streamId+partition) and `StorageNode` (currently just an address, but can be extended later).

### DIFF
--- a/src/stream/StorageNode.js
+++ b/src/stream/StorageNode.js
@@ -1,0 +1,9 @@
+export default class StorageNode {
+    constructor(address) {
+        this._address = address
+    }
+
+    getAddress() {
+        return this._address
+    }
+}
--- a/src/stream/StreamPart.js
+++ b/src/stream/StreamPart.js
@@ -1,7 +1,7 @@
 export default class StreamPart {
     constructor(streamId, streamPartition) {
         this._streamId = streamId
-        this.streamPartition = streamPartition
+        this._streamPartition = streamPartition
     }
 
     static fromStream({ id, partitions }) {
@@ -17,6 +17,6 @@ export default class StreamPart {
     }
 
     getStreamPartition() {
-        return this.streamPartition
+        return this._streamPartition
     }
 }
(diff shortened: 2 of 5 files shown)


harbu (approved):

src/stream/StreamPart.js
```diff
@@ -0,0 +1,22 @@
+export default class StreamPart {
+    constructor(streamId, streamPartition) {
+        this._streamId = streamId
+        this.streamPartition = streamPartition
```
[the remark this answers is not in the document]
teogeb: Fixed

src/stream/StreamPart.js
```diff
@@ -0,0 +1,22 @@
...
+        const result = []
+        for (let i = 0; i < partitions; i++) {
+            result.push(new StreamPart(id, i))
+        }
+        return result
+    }
+
+    getStreamId() {
+        return this._streamId
+    }
```
timoxley, line 17: These getters are very "Java style", seems weird, IMO we should probably just expose these as `this.streamId`, `this.streamPartition` or even `this.id` `this.partition`.

No need for getters.

src/stream/StreamPart.js
```diff
@@ -0,0 +1,22 @@
...
+        this._streamPartition = streamPartition
+    }
+
+    static fromStream({ id, partitions }) {
+        const result = []
+        for (let i = 0; i < partitions; i++) {
+            result.push(new StreamPart(id, i))
+        }
+        return result
+    }
```
timoxley, line 13: This is weirdly named, seems like it should be plural at least. `fromX` sounds like it should create a `StreamPart` instance from X, but this gives an array of `StreamPart` instances.

`getStreamPartsFromStream` is a bit long but maybe clearer?

src/stream/StorageNode.js
```diff
@@ -0,0 +1,9 @@
+export default class StorageNode {
+    constructor(address) {
+        this._address = address
+    }
+
+    getAddress() {
+        return this._address
+    }
+}
```
timoxley, line 9: Yeah this seems like it shouldn't be a class, or even an object… it's just a string?

timoxley (commented):
@teogeb missed this yesterday but added some comments.

PR #199, opened 2021-02-16
NET-177: Add TypeScript support

teogeb:
Added basic TypeScript support to the project. Converted most of the public methods to TypeScript.

TODO:
- many methods are currently marked with `Todo` meta type: need to define valid types for those before 5.x is published
- annotated some JS code with `@ts-expect-error`: need to check if some of the TS warnings are actually bugs
- optionality of some method parameters has not yet defined 100%: there may be parameters that are optional, but are not marked with question mark annotation or `null|undefined`
- none of the methods/fields have been marked as private: we should set valid privacy level at least to all `StreamrClient` public methods and fields


src/StreamrClient.ts
```diff
@@ -0,0 +1,583 @@
...
+import { getUserId } from './user'
+import { Todo } from './types'
+import { StreamEndpoints, StreamListQuery } from './rest/StreamEndpoints'
+import { LoginEndpoints } from './rest/LoginEndpoints'
+import { DataUnionEndpoints, DataUnionOptions } from './rest/DataUnionEndpoints'
+import { BigNumber } from '@ethersproject/bignumber'
+import Stream, { StreamProperties } from './stream'
+import { ExternalProvider, JsonRpcFetchFunc } from '@ethersproject/providers'
+
+export interface StreamrClientOptions {
```
timoxley, line 23: This should go into `./Config.ts` and could probably be largely inferred from defaults.

src/Config.ts
```diff
@@ -39,15 +41,20 @@ export default function ClientConfig(opts = {}) {
...
+            // @ts-expect-error
             url: null, // TODO: add our default public service sidechain node, also find good PoA params below
             // timeout:
             // pollingInterval:
         },
+        // @ts-expect-error
         dataUnion: null, // Give a "default target" of all data union endpoint operations (no need to pass argument every time)
         tokenAddress: '0x0Cf0Ee63788A0849fE5297F3407f701E122cC023',
         minimumWithdrawTokenWei: '1000000', // Threshold value set in AMB configs, smallest token amount to pass over the bridge
+        // @ts-expect-error
```
timoxley, line 53: Yeah these should probably be undefined or set to some default value or fix the type to be `| null` for now, IMO better than `// @ts-expect-error`

timoxley (approved):
A few minor things, but overall lgtm 👍 good work

PR #200, opened 2021-02-16
Fix resend gapfills, improve pipeline error handling

timoxley:
* Fixes and adds tests around gaps in resends.
* Standardised and added error handling where missing in pipeline.
* Explicitly ignore specific messages. Does not use orderingutil for ignored messages.
* Fixes handling for failed gapfills.
* Uses/requires typescripted `streamr-client-protocol` 8.0.0 beta-x.

Also:
* Improves AggregatedError messages to include original message + custom message.
* Excluded gapfilling tests from "no resend" tests.
* Improve test log readability by making test messages like: `{ value: "11 of 20" }`

### DIFF
--- a/src/subscribe/OrderMessages.js
+++ b/src/subscribe/OrderMessages.js
@@ -34,7 +34,7 @@ export default function OrderMessages(client, options = {}) {
         outStream.push(orderedMessage)
     }, async (from, to, publisherId, msgChainId) => {
         if (done || !gapFill) { return }
-        debug('%d gap %o', {
+        debug('gap %o', {
             streamId, streamPartition, publisherId, msgChainId, from, to,
         })
 
--- a/src/subscribe/OrderMessages.js
+++ b/src/subscribe/OrderMessages.js
@@ -83,17 +83,6 @@ export default function OrderMessages(client, options = {}) {
             }
         },
         outStream, // consumer gets outStream
-        async function* IsDone(src) {
-            for await (const msg of src) {
-                if (!gapFill) {
-                    orderingUtil.markMessageExplicitly(msg)
-                }
-
-                orderingUtil.add(msg)
-                // note no yield
-                // orderingUtil writes to outStream itself
-            }
-        },
     ], async (err) => {
         done = true
         orderingUtil.clearGaps()
--- a/src/subscribe/OrderMessages.js
+++ b/src/subscribe/OrderMessages.js
@@ -8,11 +8,13 @@ import resendStream from './resendStream'
 
 const { OrderingUtil } = Utils
 
+let ID = 0
+
 /**
  * Wraps OrderingUtil into a pipeline.
  * Implements gap filling
  */
-let ID = 0
+
 export default function OrderMessages(client, options = {}) {
     const { gapFillTimeout, retryResendAfter } = client.options
     const { streamId, streamPartition, gapFill = true } = validateOptions(options)
@@ -70,6 +72,7 @@ export default function OrderMessages(client, options = {}) {
         // no gaps are pending
         // AND
         // gaps have been filled or failed
+        // NOTE ordering util cannot have gaps if queue is empty
         if (inputClosed && orderingUtil.isEmpty()) {
             outStream.end()
         }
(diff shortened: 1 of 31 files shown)


src/utils/PushQueue.js
```diff
@@ -259,14 +279,17 @@ export default class PushQueue {
...
-                const value = await new Promise((resolve, reject) => {
+                const deferred = new Promise((resolve, reject) => {
                     // wait for next push
                     this.nextQueue.push({
                         resolve,
                         reject,
                     })
                 })
 
+                deferred.catch(() => {}) // prevent unhandledrejection
```
teogeb, line 290: Should  the `catch` be chained to the the `deferred` object? If we just call it, it creates a new promise and doesn't change the original.

src/subscribe/OrderMessages.js
```diff
@@ -53,22 +59,44 @@ export default function OrderMessages(client, options = {}) {
             resendStreams.delete(resendMessageStream)
             await resendMessageStream.cancel()
         }
-    }, gapFillTimeout, retryResendAfter)
+    }, gapFillTimeout, retryResendAfter, gapFill ? 5 : 0)
```
teogeb, line 62: Introduce a constant for this `5`? Just to avoid magic numbers. Could also add a comment if there is some specific reason why we use this value.

src/utils/Scaffold.js
```diff
@@ -23,16 +25,23 @@ export default function Scaffold(sequence = [], _checkFn, { onDone, onChange } =
     let isDone = false
     let didStart = false
 
-    function onError(err) {
-        // eslint-disable-next-line require-atomic-updates
-        error = AggregatedError.from(error, err)
+    function collectErrors(err) {
+        try {
+            if (typeof onError === 'function') {
```
teogeb, line 30: Maybe it would be enough to check if `onError` is defined (but not to check if it is a function or not). 

Passing a non-function as `onError` parameter is a configuration error. If we need to be super-protective, we can validate the type in constructor. TypeScript will solve this anyway.

timoxley: Sure, perhaps makes sense to validate this upfront.

src/subscribe/OrderMessages.js
timoxley: Yep agreed.

IMO the problem is actually more that the function takes ~6~ 7 positional arguments.

Loose rule I try to follow is no more than 2±1 positional arguments, otherwise switch to named arguments, particularly when you have a bunch of different "options". Exceptions to rule can be made when the positions make intuitive sense or for trailing object/callback argument.
 
In a future PR I'm going to switch the ordering util to look more like this:

```js
const orderingUtil = new OrderingUtil({
  streamId,
  streamPartition,
  propagationTimeout,
  resendTimeout,
  maxGapRequests: gapFill ? 5 : 0,
})

orderingUtil
    .on('message', inOrderHandler)
    .on('gap', gapHandler)
```

> Could also add a comment if there is some specific reason why we use this value

Yeah no particular reason just that if `resendTimeout` defaults to 5s then the `maxGapRequests = 10` default means stream is blocked for 50s waiting for a single message, and that seems like an awfully long time to me.

src/utils/PushQueue.js
timoxley: This is intentional. Adding a `.catch` here prevents `node --unhandled-rejections=strict` (enabled by default in next major node version) from crashing with an `unhandledrejection` when outstanding queued promises are cancelled. i.e. these particular promises are allowed to reject quietly and be handled asynchronously.

src/utils/Scaffold.js
timoxley: Ah that's right the reason I don't do this is because then you need to decide on a system for doing runtime type validation errors, and it would want to be consistent. Easier to just do this.

src/subscribe/OrderMessages.js
timoxley: Added `maxGapRequests` as a client config option in 6e83a422.

teogeb (approved):

PR #201, opened 2021-02-17
Typescript tweaks

timoxley:
### Converted to TS

* `utils/index`
* `utils/Scaffold`
* `utils/AggregatedError`
* `publish/Signer`
* `publish/Encrypt`
* `stream/StorageNode`

### Also 

* Moved config types into Config.ts
* Programmatic "mixin" of rest endpoint methods. Two steps:  grab all the functions off a rest endpoint class prototype and attach them to the client instance, bound to the rest endpoint class instance then the method types get exposed via `interface extends`. Signatures are forwarded perfectly and any private methods will still be treated as private. They will still appear on the `StreamrClient` instance, but that's kinda just how private methods in TS work.
* DU methods throw if trying to use missing `sidechainProvider`, `factorySidechainAddress`, `factoryMainnetAddress` or `tokenAddress` config.
* Remove `config.sidechainAmbAddress` seemed unused, add `config.factorySidechainAddress`.
* Use `tsc` to build node lib instead of `webpack`. Matches change in protocol 8.x.
* Update to webpack 5 for browser build. Matches change in protocol 8.x.
* Added `npm run test-types` for straight typechecking of files in both `src` + `test`. Runs in CI after linting. Use `npm run test-types -- --watch` to watch for changes.
* Added `AggregatedError.test.ts`. Improved `AggregatedError` behaviour.
* Updated a few packages to get updated typings. Noted that `maxAge` config was never working with `quick-lru@5.x`
* Fixed difficulties with bundling/using `quick-lru` package in node/browser/jest, since it's published as straight ESM. We currently want to produce a commonJS-compatible build so we need to explicitly transpile it. Note config in https://github.com/streamr-dev/streamr-client-javascript/blob/219a596e6faa084eaea5ec921b081cfc4addf42b/jest.config.js#L175-L177 and https://github.com/streamr-dev/streamr-client-javascript/pull/201/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1R176.  Likely will need to do the same for other packages as they switch over to native ESM. Sindre's likely sooner than later, see: https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77


.eslintrc.js
```diff
@@ -43,8 +53,6 @@ module.exports = {
         '@typescript-eslint/no-shadow': 'error',
         'no-unused-vars': 'off',
         '@typescript-eslint/no-unused-vars': ['error'],
-        'no-else-return': 'off',
```
teogeb, line 46: To me it is more clear if alternative but equally important return paths are indented same way:
```
if (x) {
    return 'foo'
} else {
    return 'bar'
}

```
If this not allowed, one return path looks like "primary" path and another is a fallback:
```
if (x) {
    return 'foo
}
return 'bar'
```

.eslintrc.js
```diff
@@ -43,8 +53,6 @@ module.exports = {
...
-        'no-return-await': 'off'
```
teogeb, line 47: If we allow return awaits, it is easier to remember to use await if we later refactor the last line into a separate variable:
E.g. `return getFoo()` vs. `return await getFoo()`
refactored to
```
foo = await getFoo()
return foo
```

timoxley: Hm, not sure why you'd refactor it into a separate variable though? so it's named?

Either way it doesn't make any real difference other than introducing an additional microtask, right?

.eslintrc.js
timoxley: Meh, you could still say the bit in the `if` block is the "primary", and the `else` is the fallback. IMO when faced with a low-value stylistic choice a good deciding metric is to just pick the style which minimise the amount of indentation.

I wrote a bit about avoiding `else` blocks here: https://blog.timoxley.com/post/47041269194/avoid-else-return-early

teogeb (approved):

PR #202, opened 2021-02-19
NET-205: Simplify DataUnion interface

teogeb:
DataUnion is now an object which we can operate on. To get an instance of DU object, call `client.getDataUnion(contractAddress)`. As the object encapsulates the contract address, we don't need to specify it when we access the DU methods. 

Some methods that need a member address were previously optional, and the default value for those were the currently authenticated user. Now we need to be more explicit about the parameter and the member address is always required. Also previously there was a  `dataUnion` client option which was a default value for some DU actions. That is no longer supported.

There is a new config option `factorySidechainAddress`. Users don't need to specify this as it has default value in test and will have a default value in production once we know the address.

Member join functions: 

- renamed `joinDataUnion()` -> `join(secret?)`, if secret is given this will wait until the automatic join request has been processed
- added `isMember(memberAddress)`, returns true if the member is an active member
- `hasJoined()` removed

Withdraw functions renamed to have word "all" if the amount is not specified:

- `withdrawAll(options?)`
- `withdrawAllTo(recipient, options?)`
- `signWithdrawAllTo(recipientAddress)`
- `signWithdrawAmountTo(recipientAddress, amountTokenWei)`
- admin: `withdrawFor()` -> `withdrawAllToMember(memberAddress, options?)`
- admin: `withdrawAllToSigned(memberAddress, recipientAddress, signature, options?)`
- removed methods that returned the raw withdraw transactions removed: `getWithdrawTx()`, `getWithdrawTxTo()`, `getWithdrawTxFor()`

Public query functions:

- renamed `getDataUnionStats()` -> `getStats()`
- `getMemberStats(memberAddress)`
- removed `getMembers` as it is not supported in DU v2
- renamed `getBalance()` -> `getWithdrawableEarnings(memberAddress)`
- `getAdminFee()`
- `getAdminAddress()`
- `getVersion()`

Admin functions:

- `client.deployDataUnion(options)`, waits until the deployment is ready, supports new options: `confirmations` and `gasPrice`
- `createSecret(name?)`
- `addMembers(memberAddressList, options?)`
- renamed `kick()` -> `removeMembers(memberAddressList, options?)`
- `setAdminFee(newFeeFraction)` 

Other functions:

- not related to DU: `client.getTokenBalance()`

Public accessors for DataUnion objects:

- `getAddress()`, returns string
- `getSidechainAddress()`, returns string

Internal functions:

- `client._getDataUnionFromName({ dataUnionName, deployerAddress })`, returns DataUnion
- `dataUnion._getContract()`, returns Promise<Contract>

Future refactoring, should be done after this PR is in 5.x:

- Move many DataUnionEndpoints.ts functions to DataUnion class (and inline DataUnion methods that just call an endpoint methods)
- Move `getTokenBalance` to a separate class as it is not DU-related
- Remove caching from `fetchDataUnionMainnetAddress`


README.md
```diff
@@ -328,33 +328,32 @@ This library provides functions for working with Data Unions.
 
 TODO: check all this documentation before merging/publishing, probably some of it is out of date (DU1 era)
 
-Data union functions take a third parameter, `options`, which are either overrides to options given in the constructor, or optional arguments to the function.
-
+TODO: do we need these parameters (and `confirmations` parameter) somewhere? Either add the support, or remove from `DataUnionDeployOptions` interface (currently commented out)
```
jtakalai, line 331: yea could be a good thing I guess, goes together with gasPrice really. Leaving it out for now is no issue, can make a ticket to backlog to add it back later

package.json
```diff
@@ -87,6 +88,7 @@
   },
   "#IMPORTANT": "babel-runtime must be in dependencies, not devDependencies",
   "dependencies": {
+    "@ethersproject/abi": "^5.0.12",
```
jtakalai, line 91: was this needed in the end?

jtakalai: hmm looks like yes (abi.encode)

jtakalai (approved):
DataUnion object is good stuff! So much better structure, used to be such a megafile amoeba.

Tests are passing but not exiting due to open handles. Just `--forceExit` them and make a ticket to track down the reason later. It can be really hard to find what it is, and it's probably inside ethers.js somewhere (despite doing `await providerMainnet.removeAllListeners()` in the end of tests, `--detectOpenHandles` isn't going to help much IIRC)

timoxley (commented):
lgtm

PR #205, opened 2021-02-25
Read/write streamMessage.newGroupKey & enable key revocation

timoxley:
WIP


src/StreamrClient.ts
```diff
@@ -350,6 +350,10 @@ export class StreamrClient extends EventEmitter { // eslint-disable-line no-rede
         return this.publisher.rotateGroupKey(...args)
     }
 
+    rekey(...args: Todo) {
```
teogeb, line 353: If this is an internal method, annotate with `/** @internal */`? 

The method is called in `Encryption.test.js`, after stream permission is revoked. Should the revoking trigger the rekeying automatically, without user calling this method?

src/stream/KeyExchange.js
```diff
@@ -127,7 +146,7 @@ function GroupKeyStore({ groupKeys = new Map() }) {
         },
         clear() {
             currentGroupKeyId = undefined
-            nextGroupKey = undefined
+            nextGroupKeys.length = 0
```
teogeb, line 149: Could be simply `nextGroupKeys = []`? Also at line 167.

src/stream/KeyExchange.js
```diff
@@ -96,29 +97,47 @@ function GroupKeyStore({ groupKeys = new Map() }) {
...
-            if (nextGroupKey && nextGroupKey.id === groupKeyId) { return true }
+            if (nextGroupKeys.some((nextKey) => nextKey.id === groupKeyId)) { return true }
 
             return store.has(groupKeyId)
         },
         isEmpty() {
-            return !nextGroupKey && store.size === 0
+            return nextGroupKeys.length === 0 && store.size === 0
         },
         useGroupKey() {
```
teogeb, line 107: This method could be plural as it generates an array of `GroupKey`s?

src/stream/KeyExchange.js
timoxley: I mean, sure, either way, not sure it makes a lot of difference, I was just working with it being a `const` in `const nextGroupKeys = []`.

src/stream/KeyExchange.js
timoxley: Hm, sort of, it could probably do with some docs though.

Basically it's giving you the current + (optional) next group key.

If there's two values, it'll encrypt the message with the first, but embed the second in the stream message (encrypted with the first), then the next time you call `useGroupKey` you will get that second value in the first position, and an empty value second position, until a new group key is assigned. When a new group key is assigned, it'll use the existing key one more time, embedding the next key, then start using the new key from there on.

Something like this:
```js
setGroupKey(keyA)
useGroupKey() // => [keyA, undefined]
useGroupKey() // => [keyA, undefined]
setGroupKey(keyB)
useGroupKey() // => [keyA, keyB]
useGroupKey() // => [keyB, undefined]
useGroupKey() // => [keyB, undefined]
setGroupKey(keyC)
useGroupKey() // => [keyB, keyC]
useGroupKey() // => [keyC, undefined]
// etc
```

src/StreamrClient.ts
timoxley: not internal.

> Should the revoking trigger the rekeying automatically, without user calling this method?

Probably, IIRC it did do that in the previous implementation.

src/stream/KeyExchange.js
```diff
@@ -96,29 +97,47 @@ function GroupKeyStore({ groupKeys = new Map() }) {
...
-            if (nextGroupKey) {
-                // next key becomes current key
+            const nextGroupKey = nextGroupKeys.pop()
+            // first message
+            if (!currentGroupKeyId && nextGroupKey) {
```
teogeb, line 110: To me this would be more readable if there was only one check in each `if` statement and some else statements. It would help to see the overall structure of the method (that we do exactly one of four possible things). 
```
if (nextGroupKey) {
    storeKey(nextGroupKey)
    if (currentGroupKeyId) {
        // key changed
    } else {
        // first message
    }
} else {
    if (currentGroupKeyId) {
        // ...
    } else {
        // generate & use key if none already set
    }
}
```

src/stream/KeyExchange.js
teogeb: On possibility would be to introduce some helper data type, like `GroupKeyQueue`. It would encapsulate this e.g. max length handling.

src/StreamrClient.ts
teogeb: If public, could update the type definition: `rekey(streamId: string)`

src/stream/KeyExchange.js
timoxley: I have a strong distaste for following logic through `if`/`else` statements, and even more so nested `if`/`else`.

What do you think about a `switch` statement instead?

```js
switch (true) {
    // first message
    case (!currentGroupKeyId && nextGroupKey): {
        return […]
    }
    // key changed
    case (currentGroupKeyId && nextGroupKey): {
        return […]
    }
    // generate & use key if none already set
    case (!currentGroupKeyId): {
        return […]
    }
    default: {
        return […]
    }
}
```

src/StreamrClient.ts
timoxley: Done in e13e8a1a

src/subscribe/index.ts
```diff
@@ -29,28 +29,34 @@ export class Subscription extends Emitter {
...
-    _onDone: Todo
+    _onDone: ReturnType<typeof Defer>
     /** @internal */
-    _onFinally: Todo
+    _onFinally
     /** @internal */
-    pipeline: Todo
+    pipeline
     /** @internal */
-    msgStream: Todo
```
timoxley, line 44: @teogeb FYI as of TypeScript 4.0, TS can can infer class members from usage in the constructor, so no need to list them all out with `Todo`/`any`, in most cases it can deduce the type exactly, without any annotation at all: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#class-property-inference-from-constructors.

src/stream/KeyExchange.js
```diff
@@ -537,9 +572,9 @@ export function SubscriberKeyExchange(client, { groupKeys = {} } = {}) {
     })
 
     async function getGroupKey(streamMessage) {
```
timoxley, line 574: `getGroupKey` definitely has a confusing name now though, since it returns an array of keys, one of which should be the one you're after, or an empty array.

src/stream/KeyExchange.js
teogeb: Nice, that communicates the structure cleverly! 👍

src/subscribe/index.ts
teogeb: Yeah, true. Sometimes I put the types in anyway as it makes easier to read the code: you can just check the class variables and their types by reading the source code. But IDE hover hints help, if we skip that part 😊

PR #207, opened 2021-03-01
Types for authFetch

teogeb:
Type annotations for authFetch endpoints (and `utils.test.ts`).


src/rest/StreamEndpoints.ts
```diff
@@ -248,7 +273,7 @@ export class StreamEndpoints {
         })
 
         // Send data to the stream
-        return authFetch(
+        await authFetch(
```
jtakalai, line 276: why not return?

jtakalai (approved):
Haha TS "laid-back" typing is really a breath of fresh air: sure, let the caller decide what the return type is; if they get it wrong, they get to suffer, whatever :D

src/rest/StreamEndpoints.ts
teogeb: Because the success response `POST /​streams/​id/​data` is always empty. So it is better to return void than an empty object. Error response rejects the promise.

PR #209, opened 2021-03-02
DataUnion types

teogeb:
Types for `getAdminAddress`, `setAdminFee`, `executeWithdraw`. After these modifications, and PR https://github.com/streamr-dev/streamr-client-javascript/pull/207 all public methods of `DataUnion` are annotated with valid types.

Remove `deployTxReceipt` field which was not used


src/dataunion/DataUnion.ts
```diff
@@ -296,7 +296,7 @@ export class DataUnion {
         return +adminFeeBN.toString() / 1e18
     }
 
-    async getAdminAddress(): Promise<Todo> {
+    async getAdminAddress(): Promise<string> {
```
jtakalai, line 299: did we have the EthereumAddress type alias now?

src/dataunion/DataUnion.ts
```diff
@@ -446,14 +446,14 @@ export class DataUnion {
     /**
      * Admin: set admin fee (between 0.0 and 1.0) for the data union
      */
-    async setAdminFee(newFeeFraction: number): Promise<Todo> {
+    async setAdminFee(newFeeFraction: number): Promise<void> {
```
jtakalai, line 449: why not return the receipt?

jtakalai (approved):

src/dataunion/DataUnion.ts
teogeb: Fixed

src/dataunion/DataUnion.ts
teogeb: Fixed

PR #211, opened 2021-03-02
Fix connection.test.js

teogeb:


src/Connection.ts
```diff
@@ -275,22 +285,36 @@ const DEFAULT_MAX_RETRIES = 10
...
+        return Promise.all([...openSockets].map(async (socket: Todo) => {
             socket[FORCE_CLOSED] = true // eslint-disable-line no-param-reassign
             return CloseWebSocket(socket).catch((err) => {
                 socket.debug(err) // ignore error
             })
         }))
     }
 
-    constructor(options = {}, client) {
+    constructor(options = {}, debug: Debug.Debugger) {
```
timoxley, line 315: Note original code worked like this:
https://github.com/streamr-dev/streamr-client-javascript/pull/206/commits/f3bae1382eb6d651d705195168ce8d71008212eb#diff-0e149db9f5c84dd5e3126b72e3cdcf3c297e6f2836d6a1b6bf0dea41cf05760fL66-L71

```js
        if (options.debug) {	
            this._debug = options.debug.extend(id)	
        } else {	
            this._debug = Debug(`StreamrClient::${id}`)	
        }
```

`debug` should be optional.

timoxley (approved):
Connection `debug` parameter should be optional but otherwise lgtm.

src/Connection.ts
```diff
@@ -275,22 +285,36 @@ const DEFAULT_MAX_RETRIES = 10
...
+    retryCount: Todo
+    wantsState: Todo
+    connectionHandles: Todo
+    step: Todo
+    socket?: Todo
+    didDisableAutoConnect?: Todo
+    isWaiting?: Todo
+    _isReconnecting: Todo
+    _backoffTimeout: Todo
+    sendID: Todo
```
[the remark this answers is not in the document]
teogeb: Yes, `onReconnecting` and `onMessage` work also ok without type definition. Marked the fields as `Todo` so that we'll define the actual type in the future (and the full benefit of the type system).

src/Connection.ts
teogeb: Changed that to an optional parameter. 

In the previous PR I marked `debug` as a required parameter because I looked like the `Connection` and `StreamrClient` debug instances are in practice linked together (in production code `Connection`'s debug was always derived from `StreamrClient`'s debug). But maybe optionality better as it makes `Connection` more independent in theory.

PR #214, opened 2021-03-11
Internal properties and methods

teogeb:
Hide internal properties and methods from the public classes. 

Currently the internals are annotated with a weak `/** @internal */ `. Some of the properties/methods could be marked with TypeScript's `private` to force the encapsulation. That can be done later, after we have converted all classes to TypeScript and the usage boundaries are clear.


timoxley (approved):
nice one. 👍 lgtm

PR #217, opened 2021-03-12
Upgrade MetaMask support

teogeb:
Fetch the account address from MetaMask using `request({ method: 'eth_requestAccounts' })`. Previously used the deprecated `selectedAddress` field.

Configure `chainId` to StreamrClient options. The parameter is required when using MetaMask sidechain signing.

**BREAKING CHANGE:** `client.getAddress()` is now async

We could be more explicit about network chainIds. Sidechain chainId could be a required parameter in StreamrClient `options` and we could also add chainId parameter to `options.mainnet` definition. The chainId validation can be enhanced in another PR later, if needed.

Note that there are currently no automated tests to check the MetaMask integration.

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -109,7 +109,7 @@ export default function ClientConfig(opts: StreamrClientOptions = {}) {
 
     const options: StrictStreamrClientOptions = {
         ...defaults,
-        ...opts,
+        ...opts, // sidechain is not merged with the defaults
         dataUnion: {
             ...defaults.dataUnion,
             ...opts.dataUnion
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -33,7 +33,7 @@ export type StrictStreamrClientOptions = {
     groupKeys: Todo
     keyExchange: Todo
     mainnet?: ConnectionInfo|string
-    sidechain?: ConnectionInfo|string
+    sidechain: ConnectionInfo & { chainId?: number } // TODO should the chainId be a required parameter?
     tokenAddress: EthereumAddress,
     dataUnion: {
         minimumWithdrawTokenWei: BigNumber|number|string
@@ -86,7 +86,10 @@ export default function ClientConfig(opts: StreamrClientOptions = {}) {
         // Ethereum and Data Union related options
         // For ethers.js provider params, see https://docs.ethers.io/ethers.js/v5-beta/api-providers.html#provider
         mainnet: undefined, // Default to ethers.js default provider settings
-        sidechain: 'https://rpc.xdaichain.com/',
+        sidechain: {
+            url: 'https://rpc.xdaichain.com/',
+            chainId: 100
+        },
         tokenAddress: '0x0Cf0Ee63788A0849fE5297F3407f701E122cC023',
         dataUnion: {
             minimumWithdrawTokenWei: '1000000', // Threshold value set in AMB configs, smallest token amount to pass over the bridge
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -33,7 +33,7 @@ export type StrictStreamrClientOptions = {
     groupKeys: Todo
     keyExchange: Todo
     mainnet?: ConnectionInfo|string
-    sidechain: ConnectionInfo & { chainId?: number } // TODO should the chainId be a required parameter?
+    sidechain: ConnectionInfo & { chainId?: number }
     tokenAddress: EthereumAddress,
     dataUnion: {
         minimumWithdrawTokenWei: BigNumber|number|string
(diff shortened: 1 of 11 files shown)


src/Ethereum.js
```diff
@@ -19,31 +19,32 @@ export default class StreamrEthereum {
...
         } else if (auth.ethereum) {
-            this._getAddress = () => {
-                if (auth.ethereum.selectedAddress) {
-                    throw new Error('no addresses connected+selected in Metamask')
-                }
-                return getAddress(auth.ethereum.selectedAddress)
+            this._getAddress = async () => {
+                return auth.ethereum
+                    .request({ method: 'eth_requestAccounts' })
+                    .then((accounts) => getAddress(accounts[0])) // convert to checksum case
```
hpihkala, line 29: Prefer `await` instead of `.then`?

src/Config.ts
```diff
@@ -86,7 +86,10 @@ export default function ClientConfig(opts: StreamrClientOptions = {}) {
         // Ethereum and Data Union related options
         // For ethers.js provider params, see https://docs.ethers.io/ethers.js/v5-beta/api-providers.html#provider
         mainnet: undefined, // Default to ethers.js default provider settings
-        sidechain: 'https://rpc.xdaichain.com/',
+        sidechain: {
+            url: 'https://rpc.xdaichain.com/',
+            chainId: 100
+        },
```
timoxley, line 92: Does this need merging in `options: StrictStreamrClientOptions `?

If not maybe should add a comment explicitly saying that it shouldn't be merged so it's not added later accidentally. Otherwise probably should merge:
```typescript
        sidechain: {
            ...defaults.sidechain,
            ...opts.sidechain
        },
        ```

File: src/Ethereum.js [hunk 3]
```diff
@@ -19,31 +19,33 @@ export default class StreamrEthereum {
         if (auth.privateKey) {
             const key = auth.privateKey
             const address = getAddress(computeAddress(key))
-            this._getAddress = () => address
+            this._getAddress = async () => address
```
timoxley, line 22: 👍

src/Ethereum.js
```diff
@@ -19,31 +19,33 @@ export default class StreamrEthereum {
...
             this.getSigner = () => new Wallet(key, this.getMainnetProvider())
             this.getSidechainSigner = async () => new Wallet(key, this.getSidechainProvider())
         } else if (auth.ethereum) {
-            this._getAddress = () => {
-                if (auth.ethereum.selectedAddress) {
+            this._getAddress = async () => {
+                try {
+                    const accounts = await auth.ethereum.request({ method: 'eth_requestAccounts' })
```
timoxley, line 28: Should probably cache this?

i.e. add `let address` in constructor

then
```js
this._getAddress = async () => {
            if (address) { return address }

            try {
                const accounts = await auth.ethereum.request({ method: 'eth_requestAccounts' })
                address = getAddress(accounts[0]) // convert to checksum case
                return address
            } catch {
                throw new Error('no addresses connected+selected in Metamask')
            }
        }
```

Though this doesn't prevent duplicate outstanding initial calls while `await auth.ethereum.request({ method: 'eth_requestAccounts' })` is pending, so perhaps wrap this in `pOne` or `CacheAsyncFn` from `src/utils/index.ts`

src/Ethereum.js
```diff
@@ -19,31 +19,33 @@ export default class StreamrEthereum {
...
-                // chainId is required for checking when using Metamask
                 if (!options.sidechain || !options.sidechain.chainId) {
                     throw new Error('Streamr sidechain not configured (with chainId) in the StreamrClient options!')
                 }
 
                 const metamaskProvider = new Web3Provider(auth.ethereum)
                 const { chainId } = await metamaskProvider.getNetwork()
                 if (chainId !== options.sidechain.chainId) {
-                    throw new Error(`Please connect Metamask to Ethereum blockchain with chainId ${options.sidechain.chainId}`)
+                    throw new Error(`Please connect Metamask to Ethereum blockchain with chainId ${options.sidechain.chainId}: current chainId is ${chainId}`)
```
timoxley, line 48: Can we fix max line length?

timoxley (approved):
Some minor suggestions but otherwise lgtm 👍

src/Ethereum.js
timoxley: Hm, though now I wonder what happens, and what should happen, if user does something like log out of metamask or switches metamask accounts after this stuff has been read by the client. Perhaps there's some "account changed" events or something we should be hooking into. Something to look into separately rather than in this ticket though.

src/Config.ts
teogeb: It should not merge

src/Ethereum.js
teogeb: Yes, account changes are not implemented yet. 

If the client has already fetched the address from MetaMask, the subsequent call to client.getAddress() resolve automatically to the selected accounts without any UI interaction until clients disconnects that account. So in that sense no caching is needed.

src/Ethereum.js
teogeb: Fixed

PR #218, opened 2021-03-12
Public type definitions

teogeb:
Define valid types for public API methods

Also small bug fix for parsing stream definition options (e.g. when filtering subscriptions with `getSubscription(opts)`)

### DIFF
--- a/src/StreamrClient.ts
+++ b/src/StreamrClient.ts
@@ -23,13 +23,15 @@ import { StreamPartDefinition } from './stream'
 // TODO get metadata type from streamr-protocol-js project (it doesn't export the type definitions yet)
 export type OnMessageCallback = MaybeAsync<(message: any, metadata: any) => void>
 
-// TODO check if these are correct
-export interface SubscribeOptions {
-    resend?: boolean
+export type ResendOptions = {
     from?: { timestamp: number, sequenceNumber?: number }
     to?: { timestamp: number, sequenceNumber?: number }
     last?: number
 }
+
+export type SubscribeOptions = {
+    resend?: ResendOptions
+} & ResendOptions
 
 interface MessageEvent {
     data: any
--- a/src/StreamrClient.ts
+++ b/src/StreamrClient.ts
@@ -374,12 +374,12 @@ export class StreamrClient extends EventEmitter {
         return task
     }
 
-    enableAutoConnect(...args: Todo) {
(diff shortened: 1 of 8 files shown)


src/StreamrClient.ts
```diff
@@ -18,21 +18,29 @@ import { DataUnion, DataUnionDeployOptions } from './dataunion/DataUnion'
...
 // TODO get metadata type from streamr-protocol-js project (it doesn't export the type definitions yet)
 export type OnMessageCallback = MaybeAsync<(message: any, metadata: any) => void>
 
+// TODO check if these are correct
+export interface SubscribeOptions {
+    resend?: boolean
+    from?: { timestamp: number, sequenceNumber?: number }
+    to?: { timestamp: number, sequenceNumber?: number }
+    last?: number
+}
```
timoxley, line 32: `resend` is not a boolean.

basically, resend options can either be nested or top-level in the config.

```typescript
type ResendOptions = {
    from?: { timestamp: number, sequenceNumber?: number }
    to?: { timestamp: number, sequenceNumber?: number }
    last?: number
}

type SubscribeOptions = {
    resend?: ResendOptions
} & ResendOptions
```

test/unit/StreamUtils.test.ts
```diff
@@ -0,0 +1,50 @@
+import { Stream } from '../../src/stream'
+import { validateOptions } from '../../src/stream/utils'
+
+describe('Stream utils', () => {
```
timoxley, line 4: Good

timoxley (changes requested):
resend options type is not right, but otherwise lgtm

src/subscribe/index.ts
```diff
@@ -126,6 +126,7 @@ export class Subscription extends Emitter {
         return this.pipeline.throw(...args)
     }
 
+    // TODO should we expose this to the user as no-args method?
```
teogeb, line 129: @timoxley How about this `subscription.unsubscribe()`? Is there some cases where user would this with parameters? At least internally we can call this with an `error` parameter, if I understood correctly. (Does that error parameter have some specific type?)

src/StreamrClient.ts
teogeb: Fixed

timoxley (approved):
👍 lgtm, good stuff.

PR #219, opened 2021-03-12
DU2 Docs update

fonty1:
Quick first pass at this, I expect to iterate on it next week. Docs created by `typedoc`

- Includes the factory addresses in the readme
- Typescript generated docs


README.md
```diff
@@ -326,18 +326,18 @@ All the below functions return a Promise which gets resolved with the result.
 
 This library provides functions for working with Data Unions. To get a DataUnion instance, call `client.getDataUnion(address)`. To deploy a new DataUnion, call `deployDataUnion(options)`
 
-TODO: All `options`-parameters should be documented (see TypeScript interfaces for the definitions)
+See `/docs` folder in this repo for DU `options` parameters.
 
 These DataUnion-specific options are used from `StreamrClient` options:
 | Property                            | Default                                                | Description                                                                                                      |
 | :---------------------------------- | :----------------------------------------------------- | :--------------------------------------------------------------------------------------------------------------- |
-| tokenAddress                        | 0x0Cf0Ee637<br>88A0849fE52<br>97F3407f701<br>E122cC023 | Token used by the DU                                                                                             |
```
teogeb, line 334: Users won't typically set any of these parameters. Now when we have the generated docs, this section is maybe redundant. If we want to keep it, we could maybe create a separate page about these corner case details so that the overall usage is easily understandable.

timoxley (approved):
How are people supposed to view these?
We should probably publish these using https://pages.github.com/

timoxley (changes requested):
@fonty1 can you add the doc generator to devDependencies and make an npm script like `docs`, so this process is repeatable?

README.md
fonty1: This is true, I'll remove them. I would like to have a reference page somewhere.. I'll add a linear ticket for that.

fonty1: https://github.com/streamr-dev/streamr-client-javascript/pull/219/commits/3455171e35fcd63e97c3ae3f3cbf9f64f513f7a2

fonty1:
> How are people supposed to view these?
> We should probably publish these using https://pages.github.com/

For now they run open these locally in their browser. GitHub Pages is a good idea, let's take a look at that next week. For now the script is only pointing to the Data Union docs, it might be useful to point to the rest of the client, but I need some guidance on entry points for that.

> @fonty1 can you add the doc generator to devDependencies and make an npm script like docs, so this process is repeatable?
https://github.com/streamr-dev/streamr-client-javascript/pull/219/commits/9478021b44b3e569173414f7495e11113487b6e2

timoxley (approved):
Great stuff. Thanks @fonty1

timoxley:
@fonty1 merged and set up github pages: https://streamr-dev.github.io/streamr-client-javascript/

Added links to this in readme: https://github.com/streamr-dev/streamr-client-javascript/commit/62c0fae8d9ee04db2a71cc833c448d61ff7a1cf3

Expanded generated docs to include config & streamr client, also updated DU docs in a PR here: https://github.com/streamr-dev/streamr-client-javascript/pull/221

PR #220, opened 2021-03-12
Add sendToMainnet option to DU API.

timoxley:
Begin adding sendToMainnet option to DU api. WIP.


jtakalai (approved):
I wrote most of that so I won't be able to really review it :D if tests pass and it's good for you, it's good for me.

timoxley:
Thanks for getting this across the line @teogeb @jtakalai 👍

PR #221, opened 2021-03-13
Auto-generate Client API Docs

timoxley:
Current auto-generated docs only include DataUnion docs.
This PR includes auto-generated docs for the rest of top-level client APIs.
Also added some minor annotation changes to expose some things which shouldn't be `@internal` e.g. `client.resend`, and started giving descriptions to config properties.
Some classes and methods have the `@category Important` annotation in order to highlight them in the docs.

This PR also includes updated DU docs, due to changes in #220 

----

New pages look something like this:

### Entry

image

### `StreamrClient`
image

### `Config`
image


fonty1 (approved):
Excellent! LGTM, I suppose the GitHub Pages will update every time the `/docs` folder is updated on the master branch?

timoxley:
>  I suppose the GitHub Pages will update every time the /docs folder is updated on the master branch?

Correct 👍

jtakalai (approved):
LGTM, though maybe should docs be generated and published as a Github Action, instead of pushed to source control?

timoxley:
>  though maybe should docs be generated and published as a Github Action, instead of pushed to source control?

Good point, that would be better.

timoxley:
Going to add the docs in a new workflow. Merging so Actions picks up the new workflow.

timoxley:
Hm, Actions UI wasn't showing the new workflow, but now it is. ok. Cool. Will open new PR to iterate on that.

PR #223, opened 2021-03-15
Validate client's Ethereum address options

teogeb:
If user overrides some option in `StreamrClientOptions`, we validate that value already in the constructor. Using the option is therefore easier in the code, as we know that the value is always valid.

In this PR we validate all overriden Ethereum addresses. **Open question**: Should we validate also other overriden fields this way? E.g. `restUrl`, `maxPublishQueueSize`...


test/unit/Config.test.ts
```diff
@@ -1,16 +1,43 @@
...
+            'dataUnion.templateSidechainAddress'
+        ]
+        for (const propertyPath of propertyPaths) {
+            it(propertyPath, () => {
+                const errorMessage = `${propertyPath} is not a valid Ethereum address`
+                expect(() => createClient(propertyPath, 'invalid-address')).toThrow(errorMessage)
+                expect(() => createClient(propertyPath, undefined)).toThrow(errorMessage)
+                expect(() => createClient(propertyPath, null)).toThrow(errorMessage)
+                expect(() => createClient(propertyPath, '0x1234567890123456789012345678901234567890')).not.toThrow()
+            })
```
timoxley, line 29: nice 👍

test/unit/Config.test.ts
```diff
@@ -1,16 +1,43 @@
...
         }
     })
-}
 
-describe('Config', () => {
     describe('private key', () => {
+        const createClient = (privateKey: BytesLike) => {
```
timoxley, line 34: I get why you did this but feel like having two functions with the same name but different signatures isn't great. Maybe rename one of them or make it so one function can handle both cases?

src/dataunion/Contracts.ts
```diff
@@ -293,15 +289,13 @@ export class Contracts {
...
-        validateAddress('StreamrClient factoryMainnetAddress', this.factoryMainnetAddress)
-
         if (await mainnetProvider.getCode(this.factoryMainnetAddress) === '0x') {
             throw new Error(
                 `Data union factory contract not found at ${this.factoryMainnetAddress}, check StreamrClient.options.dataUnion.factoryMainnetAddress!`
             )
         }
 
-        const factoryMainnet = new Contract(this.factoryMainnetAddress!, factoryMainnetABI, mainnetWallet)
+        const factoryMainnet = new Contract(this.factoryMainnetAddress, factoryMainnetABI, mainnetWallet)
```
timoxley, line 298: 👍

src/StreamrClient.ts
```diff
@@ -416,12 +416,8 @@ export class StreamrClient extends EventEmitter {
      */
     async getTokenBalance(address: EthereumAddress): Promise<BigNumber> {
         const { tokenAddress } = this.options
-        if (!tokenAddress) {
-            throw new Error('StreamrClient has no tokenAddress configuration.')
-        }
```
timoxley, line 421: 👍

src/Config.ts
```diff
@@ -57,8 +60,35 @@ export type StreamrClientOptions = Partial<Omit<StrictStreamrClientOptions, 'dat
...
+    // - other optional Ethereum address (if there will be some)
+    // - other overriden options (e.g. regexp check that "restUrl" is a valid url)
+    validateOverridedEthereumAddresses(opts, [
+        'streamrNodeAddress',
+        'tokenAddress',
+        'tokenSidechainAddress',
+        'dataUnion.factoryMainnetAddress',
+        'dataUnion.factorySidechainAddress',
+        'dataUnion.templateMainnetAddress',
+        'dataUnion.templateSidechainAddress'
```
timoxley, line 89: Neat, as-is this is an improvement, but note this will silently stop working if a config is renamed or if new EthereumAddress config items are added. Developer always needs to remember to update this list of paths. A shame since we do already have the type information. 

I wonder if we should do something like I described in slack, where we have an unsafe type at API boundaries, then a safe type internally and the only way to get from unsafe to safe type is to pass through the `isAddress` validator.

timoxley (approved):
Made a recommendation to use `lodash/*` instead of `lodash.*` here: https://github.com/streamr-dev/streamr-client-javascript/pull/225,
and added a few comments, but otherwise lgtm 👍 great.

test/unit/Config.test.ts
teogeb: Renamed one of the methods. The method "overloading" was quite clear from the context of each test. As all tests in that `describe` block create the client same way, short method names would be enough to understand what is happening.

PR #226, opened 2021-03-15
Split main CI workflow into test-build & test-code workflows.

timoxley:


timoxley:
Can't test without merging to master :/

PR #227, opened 2021-03-16
NET-193: Storage config propagation

teogeb:
In https://github.com/streamr-dev/broker/pull/211 we will update Broker to store only streams which are explicitly added to that StorageNode. There is only one StorageNode in each environment (production/dev). **NOTE: do not merge this before that Broker PR is in master.**

Modified `addToStorageNode` to wait until the information has been propagated to the Broker. At that point, we can be sure that any published data will be stored. To enable the propagation check, new client options were added: `storageNode.address` and `storageNode.url`.

Updated also all relevant  integration tests: if a test does a resend for some stream, that stream must have been added to a StorageNode.


test/integration/storageNode.js
```diff
@@ -0,0 +1,23 @@
...
+        return false
+    } else {
+        throw new Error('Unable to fetch stream storage status')
+    }
+}
+
+export const addToStorageNode = async (stream) => {
+    await stream.addToStorageNode(STORAGE_NODE_ADDRESS)
+    await waitForCondition(() => isStreamStoredInStorageNode(stream.id))
+}
```
timoxley, line 23: I wonder if this `waitForCondition` part should be part of `stream.addToStorageNode`? Seems like a useful thing for consumers to know when the change is actually propagated. Or perhaps the backend shouldn't respond until `isStreamStoredInStorageNode` would be true.

Additionally maybe `stream.isStoredInStorageNode()` would be useful, either as-is and/or in addition to something like `stream.waitForStreamStoredInStorageNode()`

test/integration/storageNode.js
```diff
@@ -0,0 +1,23 @@
...
+
+const isStreamStoredInStorageNode = async (streamId) => {
+    const url = `${BROKER_URL}/api/v1/streams/${encodeURIComponent(streamId)}/storage/partitions/0`
+    const response = await fetch(url)
+    if (response.status === 200) {
+        return true
+    } else if (response.status === 404) {
+        return false
+    } else {
+        throw new Error('Unable to fetch stream storage status')
```
harbu, line 16: Maybe something like:
```js
throw new Error(`Unexpected response code ${response.status} when fetching stream storage status`)
```
?

harbu (approved):

PR #228, opened 2021-03-16
Add husky & commitlint to enforce conventional-commit messages

timoxley:
This adds a commit message git hook that should "enforce" the `type(scope):` format for commit messages.

There are other rules but I've [relaxed them a bunch](https://github.com/streamr-dev/streamr-client-javascript/compare/conventional-commits?expand=1#diff-2b1ae4316086c0ebb213227790133a792300e1cef03b0afc067b23d3399ea5ca) and also set it [not run during rebases](https://github.com/streamr-dev/streamr-client-javascript/compare/conventional-commits?expand=1#diff-292bac03df29207ee254ba91cfc30951b10c50c0a9fff3a0768778cebe5f4c4aR4-R6). 

If you find any of these rules annoying, please let me know and we can just disable it. 

If there's a problem with the validation and you *really* need to skip this hook, you can always run your `git` commands with the `--no-verify` flag, e.g. 

```
git commit --no-verify 
```

Currently this is not validated in CI, local only.

Open to expanding the list of types or maybe enforcing list of scopes.
~~Possible this may cause issues with the gh-pages action, since it's writing commits to the `gh-pages` branch.~~ update: Fixed by adding additional types.


teogeb (approved):

fonty1 (approved):
Any thoughts on how this should be communicated with 3rd party devs that want to commit a fix. Perhaps a 'Contributing' section in the readme docs? I can do that in a separate PR if you'd like

jtakalai (approved):
Interesting stuff

timoxley:
> Perhaps a 'Contributing' section in the readme docs?

@fonty1 Great idea. Once this PR lands, the tooling will enforce the commit message shape, but a section about formatting messages for contributors would be good in addition to this. 

@fonty1 can you draft up a `CONTRIBUTING.md` file for the project root, linked to by the `README.md`? Using a `CONTRIBUTING` file means it would start appearing as a link in Issues/PRs i.e. https://docs.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

PR #234, opened 2021-03-22
Fix: DU withdraw amount as string

teogeb:
Bug fix for `DataUnion.signWithdrawAmountTo()`. 

The `amountTokenWei` parameter  is typed as` BigNumber|number|string`, but it if the value is a string the method throws an error: `invalid hex string (argument="value", value="...", code=INVALID_ARGUMENT, version=bytes/5.0.10)`. Fixed the parameter handling by converting the given value to a `BigNumber`. 

Fixed also a TypeScript warning by reading the value from `BigNumber` via `toHexString()` method (which is what the previous implementation did implicitly).

### DIFF
--- a/test/integration/dataunion/signature.test.ts
+++ b/test/integration/dataunion/signature.test.ts
@@ -75,7 +75,7 @@ describe('DataUnion signature', () => {
         expect(isValid3).toBe(true)
     }, 100000)
 
-    it('create signatute', async () => {
+    it('create signature', async () => {
         const client = new StreamrClient({
             auth: {
                 privateKey: '0x1111111111111111111111111111111111111111111111111111111111111111'
--- a/test/integration/dataunion/signature.test.ts
+++ b/test/integration/dataunion/signature.test.ts
@@ -84,10 +84,11 @@ describe('DataUnion signature', () => {
         const dataUnion = client.getDataUnion('0x2222222222222222222222222222222222222222')
         const to = '0x3333333333333333333333333333333333333333'
         const withdrawn = BigNumber.from('4000000000000000')
-        // 0.005 tokens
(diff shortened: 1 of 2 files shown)


test/integration/dataunion/signature.test.ts
```diff
@@ -74,4 +74,21 @@ describe('DataUnion signature', () => {
...
+        const dataUnion = client.getDataUnion('0x2222222222222222222222222222222222222222')
+        const to = '0x3333333333333333333333333333333333333333'
+        const withdrawn = BigNumber.from('4000000000000000')
+        const amounts = [5000000000000000, '5000000000000000', BigNumber.from('5000000000000000')]
+        // eslint-disable-next-line no-underscore-dangle
+        const signaturePromises = amounts.map((amount) => dataUnion._createWithdrawSignature(amount, to, withdrawn, client.ethereum.getSigner()))
+        const actualSignatures = await Promise.all(signaturePromises)
+        const expectedSignature = '0x5325ae62cdfd7d7c15101c611adcb159439217a48193c4e1d87ca5de698ec5233b1a68fd1302fdbd5450618d40739904295c88e88cf79d4241cf8736c2ec75731b' // eslint-disable-line max-len
+        expect(actualSignatures.every((actual) => actual === expectedSignature))
+    })
```
timoxley, line 93: I feel like we should have a test around `signWithdrawAmountTo` rather than/in addition to `_createWithdrawSignature`, since `_createWithdrawSignature` is just implementation detail for `signWithdrawAmountTo`.

timoxley (approved):
I'd like to see the test around the exposed interface `signWithdrawAmountTo`, if that's not too much of a PITA to set up, but otherwise LGTM.

timoxley:
Merging this to get it into 5.2 release. My comment on the test is not blocking.

PR #235, opened 2021-03-22
NET-141: Stream ids in Network API should be case-insensitive

bearni95:


timoxley:
@bearni95 this is no longer going ahead, right?

bearni95:
nopes

timoxley:
Closed, stale.

Issue #233, opened 2021-03-21
JavaScript heap out of memory


aseyfpour:
Hi

I'm trying to subscribe to a stream using resend option. Everything goes well but after a while, it crashes.

streamr-client: 5.1.0

Stacktrace:
```
<--- Last few GCs --->

[2968:0000021AA9CE65C0]   362454 ms: Mark-sweep 2035.3 (2073.9) -> 2029.6 (2059.1) MB, 855.3 / 0.0 ms  (+ 862.9 ms in 233 steps since start of marking, biggest step 36.1 ms, walltime since start of marking 1943 ms) (average mu = 0.186, curre
nt mu = 0.116)[2968:0000021AA9CE65C0]   364209 ms: Mark-sweep 2036.5 (2059.1) -> 2025.8 (2053.1) MB, 1265.2 / 0.0 ms  (+ 388.4 ms in 109 steps since start of marking, biggest step 47.6 ms, walltime since start of marking 1755 ms) (average mu
 = 0.139, current mu = 0.058

<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 00007FF69653770D]
    1: StubFrame [pc: 00007FF6964E70DC]
Security context: 0x02f6497808d1 <JSObject>
    2: /* anonymous */ [000003C11D8E3A01] [D:\Streamr\data-receiver\node_modules\streamr-client\src\subscribe\OrderMessages.js:~33] [pc=0000011ACAA8E7DA](this=0x003d5c513709 <EventEmitter map = 0000032521B25699>,0x022f89d542a1 <MessageRef ma
p = 0000032521B21549>,0x0199f1059a39 <MessageRef map = 0000032521B21549>,0x03...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF6959246CF napi_wrap+113583
 2: 00007FF6958CF7C6 v8::base::CPU::has_sse+66646
 3: 00007FF6958D05C6 v8::base::CPU::has_sse+70230
 4: 00007FF6960E42AE v8::Isolate::ReportExternalAllocationLimitReached+94
 5: 00007FF6960CC381 v8::SharedArrayBuffer::Externalize+833
 6: 00007FF695F988CC v8::internal::Heap::EphemeronKeyWriteBarrierFromCode+1436
[...]

PR #236, opened 2021-03-25
NET-229: Fix Memleak(s)

timoxley:
Fixes for #233.

Found two definite leaks, and a few minor potential leaks.
Resend's `last` option limits to 10,000 messages, but no such limit for resend `from`/`to`.
Resend limits for `from`/`to` with some kind of pagination like this stale PR would probably help a lot: https://github.com/streamr-dev/broker/pull/190

----

CI failure due to production resends being broken.

Might have to figure out a good way to replay large datasets in test environment, it's quite possible my testing of this yesterday caused production resends to fall over. Of course this shouldn't happen at all, and we probably should do something to break up resends into smaller chunks, for the benefit of both producer and consumer.


src/utils/PushQueue.ts
```diff
@@ -247,16 +236,23 @@ export default class PushQueue<T> {
     async _cleanup() {
         this.finished = true
         const { error } = this
-        for (const p of this.nextQueue) {
+        const queue = this.nextQueue
+        this.error = undefined
+        this.nextQueue = []
+        this.pending = 0
+        this.buffer.length = 0
+        while (queue.length) {
```
teogeb, line 244: Could be `this.buffer = []`, and `while (queue.length > 0)`

src/utils/PushQueue.ts
```diff
@@ -247,16 +236,23 @@ export default class PushQueue<T> {
...
+            const p = queue.shift()
+            if (!p) { continue }
```
teogeb, line 246: Compare explicitly to `undefined`, so that the queue can contain falsy values.

teogeb (approved):

src/utils/PushQueue.ts
timoxley: Good thought, though the "queue" here doesn't contain values, it's a queue of promises generated by calls to `next()` while the current async `next` is inflight i.e. should never contain anything but `Defer` promises. The `this.buffer` is what contains the buffered values, though given this class is called `PushQueue` I can see the confusion. I'll rename the `queue` variable to `nextQueue`

timoxley:
CI failure seems unrelated, merging so we can put out a new, fixed 5.x release.

PR #239, opened 2021-03-30
ETH-81: Expose transportSignatures for an alternative withdraw flow

jtakalai:
New withdraw functionality.

Alpha version is already in NPM, this PR is just adding tests. After that it should be good for master.

### DIFF
--- a/src/dataunion/DataUnion.ts
+++ b/src/dataunion/DataUnion.ts
@@ -143,7 +143,10 @@ export class DataUnion {
 
     /**
      * Withdraw all your earnings
-     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false, ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost), ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves, ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false
+     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false,
+     *          ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false,
+     *          ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves,
+     *          ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost)
      */
     async withdrawAll(options?: DataUnionWithdrawOptions) {
         const recipientAddress = await this.client.getAddress()
@@ -177,7 +180,10 @@ export class DataUnion {
 
     /**
      * Withdraw earnings and "donate" them to the given address
-     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false, ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost), ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves, ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false
+     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false,
+     *          ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false,
+     *          ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves,
+     *          ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost)
      */
     async withdrawAllTo(
         recipientAddress: EthereumAddress,
@@ -403,7 +409,10 @@ export class DataUnion {
      * Admin: withdraw earnings (pay gas) on behalf of a member
      * TODO: add test
      * @param memberAddress - the other member who gets their tokens out of the Data Union
-     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false, ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost), ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves, ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false
+     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false,
+     *          ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false,
+     *          ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves,
+     *          ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost)
      */
     async withdrawAllToMember(
         memberAddress: EthereumAddress,
@@ -433,7 +442,10 @@ export class DataUnion {
      * @param memberAddress - the member whose earnings are sent out
      * @param recipientAddress - the address to receive the tokens in mainnet
      * @param signature - from member, produced using signWithdrawAllTo
-     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false, ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost), ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves, ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false
+     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false,
+     *          ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false,
+     *          ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves,
+     *          ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost)
      */
     async withdrawAllToSigned(
         memberAddress: EthereumAddress,
@@ -565,7 +577,10 @@ export class DataUnion {
     /**
      * Template for withdraw functions
      * @private
-     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false, ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost), ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves, ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false
+     * @returns the sidechain withdraw transaction receipt IF called with sendToMainnet=false,
+     *          ELSE the message hash IF called with transportSignatures=false and waitUntilTransportIsComplete=false,
+     *          ELSE the mainnet AMB signature execution transaction receipt IF we did the transport ourselves,
+     *          ELSE null IF transport to mainnet was done by someone else (in which case the receipt is lost)
      */
     private async _executeWithdraw(
         getWithdrawTxFunc: () => Promise<ContractTransaction>,
@@ -575,7 +590,8 @@ export class DataUnion {
         const {
             pollingIntervalMs = 1000,
             retryTimeoutMs = 60000,
-            transportSignatures = !this.client.options.dataUnion.freeWithdraw, // by default, transport the signatures if freeWithdraw isn't supported by the sidechain
+            // by default, transport the signatures if freeWithdraw isn't supported by the sidechain
+            transportSignatures = !this.client.options.dataUnion.freeWithdraw,
             waitUntilTransportIsComplete = true,
             sendToMainnet = true,
         } = options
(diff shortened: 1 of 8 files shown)


jtakalai:
Data union tests are all green. Merging after a review.

src/dataunion/DataUnion.ts
```diff
@@ -449,15 +474,15 @@ export class DataUnion {
...
+    ) {
         const duSidechain = await this.getContracts().getSidechainContract(this.contractAddress)
         return duSidechain.withdrawAllToSigned(memberAddress, recipientAddress, sendToMainnet, signature)
     }
 
     /**
      * Admin: set admin fee (between 0.0 and 1.0) for the data union
      */
-    async setAdminFee(newFeeFraction: number): Promise<TransactionReceipt> {
+    async setAdminFee(newFeeFraction: number) {
```
teogeb, line 485: The return type should still be `Promise<TransactionReceipt>`?

src/dataunion/DataUnion.ts
```diff
@@ -564,30 +590,109 @@ export class DataUnion {
...
+    }
+
+    // TODO: this doesn't belong here. Transporting a message is NOT dataunion-specific and needs nothing from DataUnion.ts.
+    //       It shouldn't be required to create a DataUnion object to call this.
+    //       This belongs to the StreamrClient, and if the code is too DU-specific then please shove it back to Contracts.ts.
+    //       Division to transportMessage and Contracts.transportSignaturesForMessage is spurious, they should be one long function probably.
+    /**
+     * @returns null if message was already transported, ELSE the mainnet AMB signature execution transaction receipt
+     */
+    async transportMessage(messageHash: AmbMessageHash, pollingIntervalMs: number = 1000, retryTimeoutMs: number = 300000) {
```
teogeb, line 670: The blockchain-level code (`transportMessage` and part of `_executeWithdraw`) could be moved to `dataunion/Contracts.ts`

src/dataunion/DataUnion.ts
```diff
@@ -564,30 +590,109 @@ export class DataUnion {
...
         options: DataUnionWithdrawOptions = {}
-    ): Promise<TransactionReceipt> {
+    ): Promise<ContractReceipt | AmbMessageHash | null> {
         const {
             pollingIntervalMs = 1000,
-            retryTimeoutMs = 60000,
-            freeWithdraw = this.client.options.dataUnion.freeWithdraw,
+            retryTimeoutMs = 300000,
+            // by default, transport the signatures if freeWithdraw isn't supported by the sidechain
+            payForTransport = !this.client.options.dataUnion.freeWithdraw,
```
teogeb, line 610: Is `payForTransport` just the same concept as `freeWithdraw`, but negated? Do we need both terms?

test/integration/dataunion/withdraw.test.ts
```diff
@@ -44,14 +50,16 @@ const testWithdraw = async (
...
-    const adminClient = new StreamrClient(config.clientOptions as any)
+    const adminClient = new StreamrClient(config.clientOptions)
 
     const dataUnion = await adminClient.deployDataUnion()
     const secret = await dataUnion.createSecret('test secret')
     log(`DataUnion ${dataUnion.getAddress()} is ready to roll`)
     // dataUnion = await adminClient.getDataUnionContract({dataUnion: "0xd778CfA9BB1d5F36E42526B2BAFD07B74b4066c0"})
 
-    const memberWallet = new Wallet(`0x100000000000000000000000000000000000000012300000001${Date.now()}`, providerSidechain)
+    testWalletId += 1
```
timoxley, line 60: 👍 this is much better than `Date.now()`

test/integration/dataunion/withdraw.test.ts
```diff
@@ -149,74 +183,85 @@ const testWithdraw = async (
...
     })
 
-    for (const sendToMainnet of [true, false]) {
+    describe.each([
+        [false, true, true], // sidechain withdraw
+        [true, true, true], // self-service mainnet withdraw
+        [true, true, false], // self-service mainnet withdraw without checking the recipient account
+        [true, false, true], // bridge-sponsored mainnet withdraw
+        [true, false, false], // other-sponsored mainnet withdraw
+    ])('Withdrawing with sendToMainnet=%p, payForTransport=%p, wait=%p', (sendToMainnet, payForTransport, waitUntilTransportIsComplete) => {
```
timoxley, line 220: Neat

src/dataunion/DataUnion.ts
```diff
@@ -564,30 +590,109 @@ export class DataUnion {
...
+                log(`Waiting for balance=${balanceBefore.toString()} change (poll every ${pollingIntervalMs}ms, timeout after ${retryTimeoutMs}ms)`)
+                await until(async () => !(await getBalanceFunc()).eq(balanceBefore), retryTimeoutMs, pollingIntervalMs).catch((e) => {
+                    const msg = `Timeout: Bridge did not transport withdraw message as expected. Fix: DataUnion.transportMessage("${messageHash}")`
+                    throw e.message.startsWith('Timeout') ? new Error(msg) : e
+                })
+                return null
+            }
+
+            // instead of waiting, hand out the messageHash so that someone else might do the transport using it
+            return messageHash
```
timoxley, line 652: This return type is a bit weird `ContractReceipt | AmbMessageHash | null`, I wonder if instead of a `payForTransport` flag we should have two methods? Might be a bit of work though.

src/dataunion/DataUnion.ts
```diff
@@ -564,30 +590,109 @@ export class DataUnion {
...
+            throw new Error(`Message with hash=${messageHash} not found`)
         }
-        log(`Waiting for balance ${balanceBefore.toString()} to change`)
-        await until(async () => !(await getBalanceFunc()).eq(balanceBefore), retryTimeoutMs, pollingIntervalMs)
-        return tr
+        const messageId = '0x' + message.substr(2, 64)
+
+        log(`Checking mainnet AMB hasn't already processed messageId=${messageId}`)
+        const alreadySent = await mainnetAmb.messageCallStatus(messageId)
+        const failAddress = await mainnetAmb.failedMessageSender(messageId)
```
timoxley, line 686: Maybe could do a pass on all this to find things like this that can probably be done in parallel e.g.:

```js
const [alreadySent, failAddress] = await Promise.all([
  mainnetAmb.messageCallStatus(messageId),
  mainnetAmb.failedMessageSender(messageId),
])
```

test/integration/dataunion/withdraw.test.ts
```diff
@@ -131,13 +139,39 @@ const testWithdraw = async (
...
+    const balanceBefore = await getRecipientBalance()
     log(`Balance before: ${balanceBefore}. Withdrawing tokens...`)
 
-    const withdrawTr = await withdraw(dataUnion.getAddress(), memberClient, memberWallet, adminClient)
-    log(`Tokens withdrawn, sidechain tx receipt: ${JSON.stringify(withdrawTr)}`)
-    const balanceAfter = await getBalance(memberWallet)
+    // "bridge-sponsored mainnet withdraw" case
+    if (!options.payForTransport && options.waitUntilTransportIsComplete) {
+        log(`Adding ${recipient} to bridge-sponsored withdraw whitelist`)
+        bridgeWhitelist.push(recipient)
```
timoxley, line 154: Does `recipient` need to be `toLowerCase`?

test/integration/dataunion/withdraw.test.ts
```diff
@@ -131,13 +139,39 @@ const testWithdraw = async (
...
+
+    // test setup done, do the withdraw
+    let ret = await withdraw(dataUnion.getAddress(), memberClient, memberWallet, adminClient)
+
+    // "other-sponsored mainnet withdraw" case
+    if (typeof ret === 'string') {
+        log(`Transporting message "${ret}"`)
+        ret = await dataUnion.transportMessage(String(ret))
+    }
+    log(`Tokens withdrawn, return value: ${JSON.stringify(ret)}`)
```
timoxley, line 165: Remember you can use string formatting tokens instead of stringify e.g. 
```js
log(`Tokens withdrawn, return value: %o,  ret)
```

test/integration/dataunion/withdraw.test.ts
```diff
@@ -44,14 +50,16 @@ const testWithdraw = async (
...
+    const adminClient = new StreamrClient(config.clientOptions)
 
     const dataUnion = await adminClient.deployDataUnion()
     const secret = await dataUnion.createSecret('test secret')
     log(`DataUnion ${dataUnion.getAddress()} is ready to roll`)
     // dataUnion = await adminClient.getDataUnionContract({dataUnion: "0xd778CfA9BB1d5F36E42526B2BAFD07B74b4066c0"})
 
-    const memberWallet = new Wallet(`0x100000000000000000000000000000000000000012300000001${Date.now()}`, providerSidechain)
+    testWalletId += 1
+    const memberWallet = new Wallet(`0x100000000000000000000000000000000000000012300000000000001${testWalletId}`, providerSidechain)
```
timoxley, line 61: Hm, I do wonder if it'd avoid possible state leaking between tests if the addresses were always random? e.g. 

```js
`0x${crypto.randomBytes(32).toString('hex')}`
```

src/dataunion/DataUnion.ts
timoxley, line 670: `retryTimeoutMs: number = 300000`

5 minutes seems like a long time

timoxley (approved):
Would like it if more functions took options objects instead of positional parameters, my IDE doesn't have the nice positional naming stuff that yours does, but otherwise LGTM.

src/dataunion/DataUnion.ts
jtakalai: those are kind of... special cases of the same thing, somehow. Yes the type is super weird, but I guess the other solution would be to rewrite it so that it's exported in smaller pieces so that if you don't want the "whole thing" you can just take the pieces you want, or something. Though that's how it is now, too, more or less.

test/integration/dataunion/withdraw.test.ts
jtakalai: sure, though I think there is value in deterministic addresses too

test/integration/dataunion/withdraw.test.ts
jtakalai: it needs to eventually be, it's toLowerCased in the "simulated bridge". I think it's good to keep the checksum-case where possible.

test/integration/dataunion/withdraw.test.ts
jtakalai: interesting, I'll try it!

test/integration/dataunion/withdraw.test.ts
jtakalai: I mean, technically those numbers/keys/addresses are "state" but it's never used except as an opaque object, and it only needs to be unique. Random addresses would work just as well, that was the idea of the Date.now() pretty much. But I'll keep that randomBytes thing in mind.

test/integration/dataunion/withdraw.test.ts
jtakalai: I'll just do a complete logging cleanup now

src/dataunion/DataUnion.ts
jtakalai: ContractReceipt now, and it's in fact correctly inferred. Manually typing it to TransactionReceipt (a supertype) was in fact harmful.

src/dataunion/DataUnion.ts
jtakalai: yea but we don't know how the bridge oracles operate. 5 minutes might in fact be what it may take in production. It's not great for web sites obviously. Maybe there needs to be a more multi-step process where you can add a listener for when it's ready or smth (e.g. get back the message hash anyway, store it to localstorage, then start listening to when it's transported; if page gets refreshed or something, get that "pending hash" from localstorage and resume listening)

None of that is something I've planned as part of this ticket. I'll make a new ticket, but the feature is going to be for the front-end people, so I can't really say how urgent it would be. It would change the public interface though...

jtakalai: Made a ticket https://linear.app/streamr/issue/ETH-101/streamr-client-javascript-data-union-ethereum-utils-refactor

jtakalai: https://linear.app/streamr/issue/ETH-102/split-withdraw-with-signature-transport-more-frontend-ergonomically

src/dataunion/DataUnion.ts
jtakalai: it is. One is the property of the bridge, one is the property of the function call, it's like the opposite angle of looking at it.

But I'll still change it so that the name is the same. So freeWithdraw -> !payForTransport everywhere. It's better that way.

PR #240, opened 2021-03-30
Fix CI

timoxley:
CI seems to be failing due to:
* memory tests timing out
* Flakiness in "multiple client" tests. They pass in isolation, but not when run in suite.
* [Storage node is no longer accessible](https://github.com/streamr-dev/streamr-docker-dev/pull/102) and, 
* A 502 error connecting to ws://localhost/api/v1/ws, likely related to storage node change above.


timoxley:
Just disabling the client testing tool tests for now.
The thing doesn't give reliable results and I think it may be the tool itself that needs some reworking.

PR #242, opened 2021-04-12
Enable "stream-encrypted-exchanged-rotating-revoking-signed" tests

teogeb:
Now runs all `client-testing-tool` tests.


timoxley (approved):
lgtm

PR #248, opened 2021-05-01
fix(dataunion): Fix inverted poll timeout + interval.

timoxley:
Theory is that this should make the dataunion tests run faster.


jtakalai (approved):
good catch!

PR #250, opened 2021-05-03
Improve "until": Don't evaluate the condition twice if it's true from the beginning

jtakalai:
also fix comments

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -488,11 +488,14 @@ export async function sleep(ms: number = 0) {
     })
 }
 
+// condition could as well return any instead of boolean, could be convenient sometimes if waiting until a value is returned. Maybe change if such use case emerges.
 /**
  * Wait until a condition is true
  * @param condition - wait until this callback function returns true
  * @param timeOutMs - stop waiting after that many milliseconds, -1 for disable
  * @param pollingIntervalMs - check condition between so many milliseconds
+ * @param failedMsgFn - append the string return value of this getter function to the error message, if given
+ * @return the (last) truthy value returned by the condition function
  */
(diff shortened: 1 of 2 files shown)


timoxley (approved):
Great change, especially if `condition` is `async`.

Could add a test?

src/utils/index.ts
```diff
@@ -488,11 +488,14 @@ export async function sleep(ms: number = 0) {
     })
 }
 
+// condition could as well return any instead of boolean, could be convenient sometimes if waiting until a value is returned. Maybe change if such use case emerges.
```
timoxley, line 491: `boolean` is perfectly fine IMO.
`any` doesn't communicate that we're expecting a `truthy` or `falsey` result, and in the worst case someone just needs to do `return !!value` instead of `return value`

Also note line length too long.

src/utils/index.ts
```diff
@@ -511,6 +515,7 @@ export async function until(condition: MaybeAsync<() => boolean>, timeOutMs = 10
         }
 
         await sleep(pollingIntervalMs) // eslint-disable-line no-await-in-loop
+        wasDone = await Promise.resolve().then(condition) // eslint-disable-line no-await-in-loop
     }
-    return condition()
+    return wasDone
```
timoxley, line 520: This can never return `false` anyway right? it'll always either return `true` or `throw` with timeout

src/utils/index.ts
timoxley: Ahh I see, you mean it could return the first/last `truthy` result of calling `condition()`… interesting.

Probably don't want `any` for that, instead you'd infer the type from `condition`, something like:

```typescript
async function until<T>(condition: MaybeAsync<() => T>, …): Promise<T> => {
…
}
```

PR #254, opened 2021-05-10
NET-286: Storage node constants

teogeb:
Added constants `StorageNode.STREAMR_GERMANY` and `StorageNode.STREAMR_DOCKER_DEV`. The values are instances of `StorageNode` class.

The parameter of `addToStorageNode`, `removeFromStorageNode` and `getStreamPartsByStorageNode` can now be a `StorageNode` instance or a `string`.


timoxley (approved):
Yep this is much better. Worth updating readme? Otherwise lgtm.

PR #256, opened 2021-05-17
NET-236: Create stream by path

teogeb:
The `id` parameter of `client.createStream` can  be a path (or full stream id, as previously). It will be prefixed with a lowercased Ethereum address derived from user's private key.

E.g. 
```
client.createStream({
    id: '/foo/bar'
})
// -> creates 0xabcdeabcde123456789012345678901234567890/foo/bar
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,8 @@ See "Subscription options" for resend options
 ### Programmatically creating a stream
 
 ```js
-const stream = await client.getOrCreateStream({
-    name: 'My awesome stream created via the API',
+const stream = await client.createStream({
+    id: '/foo/bar', // or 0x1234567890123456789012345678901234567890/foo/bar or mydomain.eth/foo/bar
 })
 console.log(`Stream ${stream.id} has been created!`)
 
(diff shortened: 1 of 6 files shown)


timoxley (approved):
feels weird to pass in an `id` and get back an object with a different `id` but I guess the client itself does the same thing.

I added a test to check it wasn't possible to create a stream with another user's address as the prefix, it passed though perhaps client could detect this before it hits the backend and give a nicer error, maybe:
https://github.com/streamr-dev/streamr-client-javascript/pull/256/commits/ffbb273a300bd9383834f8d5518dcb8101a6736e

teogeb:
> feels weird to pass in an `id` and get back an object with a different `id` but I guess the client itself does the same thing.
> 
> I added a test to check it wasn't possible to create a stream with another user's address as the prefix, it passed though perhaps client could detect this before it hits the backend and give a nicer error, maybe:
> [ffbb273](https://github.com/streamr-dev/streamr-client-javascript/commit/ffbb273a300bd9383834f8d5518dcb8101a6736e)

Agree. Alternatively we could use something like `createStream({ id: createStreamId('/foo/bar') }` but it wouldn't be very readable. Maybe this is a good compromise. Backend validation is fine, but maybe error messages could be better.

PR #258, opened 2021-05-26
fix: ETH-111 add retry logic to tx.wait for sidechain transactions

jtakalai:


src/dataunion/DataUnion.ts
```diff
@@ -77,6 +77,22 @@ function getMessageHashes(tr: ContractReceipt): AmbMessageHash[] {
...
+            const msg = body.error.message
+            log('Error message: %s', msg)
+            if (msg.indexOf('ancient block sync') >= 0) {
+                log('Sleeping then trying again')
+                return sleep(10000).then(() => waitForTx(tx)) // TODO: add option for retry interval and timeout
+            }
+        }
+        throw e
+    })
+}
```
timoxley, line 94: Added fixed number of retries & retry interval here:

```typescript
type WaitForTXOptions = {
    retries?: number
    retryInterval?: number
}

async function waitForTx(tx: ContractTransaction, { retries = 3, retryInterval = 10000 }: WaitForTXOptions = {}): Promise<ContractReceipt> {
    return tx.wait().catch((err) => {
        log('Got error: %o', err)
        if (err.body) {
            const body = JSON.parse(err.body)
            const msg = body.error.message
            log('Error message: %s', msg)
            if (retries > 0 && msg.includes('ancient block sync')) {
                log('Sleeping for %dms then retrying %d more time(s).', retryInterval, retries)
                await sleep(10000)
                return waitForTx(tx, { retries: retries - 1, retryInterval })
            }
        }
        throw err
    })
}
```

timoxley: Maybe the function could be `waitForTxWithRetry` or something

timoxley: also renamed `e` to `err` so it's greppable in an emergency.

src/dataunion/DataUnion.ts
```diff
@@ -77,6 +77,22 @@ function getMessageHashes(tr: ContractReceipt): AmbMessageHash[] {
     return hashes
 }
 
+async function waitForTx(tx: ContractTransaction): Promise<ContractReceipt> {
+    return tx.wait().catch((e) => {
+        log('Got error: %o', e)
```
timoxley, line 82: Does the error message include information about which transaction failed, wondering if something like `tx.hash` should be printed here too.

timoxley (approved):
a few suggestions but otherwise lgtm

src/dataunion/DataUnion.ts
jtakalai: good point, might as well print the tx too

src/dataunion/DataUnion.ts
jtakalai: good stuff, I really don't mind if you'd simply commit them  😅  but I'll do it now then

PR #259, opened 2021-05-26
fix: strip ansi-colors to get the branch name correctly

jtakalai:
I got an error like this:
```
$ git commit
.husky/commit-msg: line 4: [: too many arguments
```
the commit went through but without the check

the reason I think is that my `git branch` adds colors, so this new sed magic now strips that ansi-coloring.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,5 +1,5 @@
 #!/bin/sh
-BRANCH_NAME=$(git branch --no-color | grep --color=never '*' | sed 's/* //')
+BRANCH_NAME=$(git symbolic-ref --short -q HEAD)
 
 if [ $BRANCH_NAME != '(no branch)' ]
 then
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,5 +1,5 @@
 #!/bin/sh
-BRANCH_NAME=$(git branch | grep '*' | sed 's/* //')
+BRANCH_NAME=$(git branch | grep '*' | sed 's/.*\*.* //')
 
 if [ $BRANCH_NAME != '(no branch)' ]
 then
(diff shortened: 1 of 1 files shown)


.husky/commit-msg
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-BRANCH_NAME=$(git branch | grep '*' | sed 's/* //')
+BRANCH_NAME=$(git branch | grep '*' | sed 's/.*\*.* //')
```
kare, line 2: I think it should be: `git branch --no-color`

timoxley: Nice one @kare

timoxley (changes requested):
if Kare's `--no-color` works then we should probably do that

jtakalai:
it was `grep` putting the color in there (highlighting the "`*`" it found)

.husky/commit-msg
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-BRANCH_NAME=$(git branch | grep '*' | sed 's/* //')
+BRANCH_NAME=$(git branch --no-color | grep --color=never '*' | sed 's/* //')
```
kare, line 2: Looks like my `git branch` solution is not the correct one. While scripting with Git plumbing commands should be used and `git branch` is not a plumbing but porcelain command.
https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain
https://git-blame.blogspot.com/2013/06/checking-current-branch-programatically.html

TLDR: `git symbolic-ref --short -q HEAD`

jtakalai: ok! maybe that one then

kare (approved):
LGTM

PR #265, opened 2021-06-02
feature: add async safeGetDataUnion

jtakalai:
make another function in order to avoid a breaking change (async)

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,11 @@ To get an existing (previously deployed) `DataUnion` instance:
 const dataUnion = client.getDataUnion(dataUnionAddress)
 ```
 
+Or to verify untrusted (e.g. user) input, use:
+```js
+const dataUnion = await client.safeGetDataUnion(dataUnionAddress)
+```
+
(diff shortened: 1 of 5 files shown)


teogeb:
Good improvement! 👍 To me it sounds that this safe behavior could be the default, and no separate method is needed. 

Are there situations where the methods of `DataUnion` object would work practice, if we get it by `DataUnion.getInstance(addr)` where the `addr` is not an address of a DU contract? 

Some methods would of course work if there is a contract which implements only part of the ABI. But should we to support that kind of use cases?

jtakalai:
> this safe behavior could be the default

it incurs an HTTP call and also makes it async. Maybe safe should be default and "uncheckedGetDataUnion" maybe the "optimized" way. Usually you get addresses from configs so they "can't be wrong", and if they're wrong (or if you're in wrong chain or whatever) you can't really recover anyway. This "safe" option is good for user-given inputs.

fonty1:
@jtakalai AFAIK this should land in the monorepo js client instead?

PR #267, opened 2021-06-14
config.js => config.ts

jtakalai:
makes adding more Ethereum/DU utils easier, makes tests DRYer


test/integration/config.ts
```diff
@@ -0,0 +1,92 @@
+import { Wallet } from 'ethers'
```
teogeb, line 1: Do the benchmark tests and other test written in JS work? 

At least `npm run benchmarks` command fails as `publish.js` JS-file imports to a `config.ts` TypeScript file. Maybe we need to convert all tests to TypeScript, or import the built JS file from `dist` folder.

teogeb:
Some tests have not yet updated? `npm run eslint` fails with errors (mostly `'clientOptions' is not defined`)

test/integration/config.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// can mint mainnet DATA tokens
+export const tokenAdminPrivateKey = '0x5e98cce00cff5dea6b454889f359a4ec06b9fa6b88e9d69b86de8e1c81887da0'
+
+export const providerSidechain = new JsonRpcProvider(clientOptions.sidechain)
+export const providerMainnet = new JsonRpcProvider(clientOptions.mainnet)
+
+export function getTestWallet(index: number, provider: JsonRpcProvider) {
+    // TODO: change to 'streamr-client-javascript' once https://github.com/streamr-dev/smart-contracts-init/pull/36 is in docker
+    const hash = id('marketplace-contracts' + (index || ''))
```
teogeb, line 51: Is the `|| ''` operation needed? As index is a non-null number that special is applied only if index is zero.

test/integration/config.ts
jtakalai: I'm unable to run any tests because of babel errors

jtakalai: but that's a good point. I'll keep the old js file and just make the ts file under different name. Good catch.

jtakalai: seems IDEA refactors don't spot that kind of problems  :x  worrying.

jtakalai:
Some unit tests all passed but the test itself still failed "Error: The action has timed out." 

Add dataunion tests pass, too, so I'm calling this a success.

PR #268, opened 2021-06-15
Wrap transferWithinContract in the smart contract

jtakalai:


jtakalai:
dataunions tests pass, looks like another success.

